PR #16, opened 2025-10-07
Update router.py to pass auth

harishsg999:
Update router.py to pass auth token to scanner


ablox-chetan:
Is there a unit test or manual test to show how this works ?

ablox-chetan (approved):

PR #25, opened 2025-10-15
fix: Handle unavailable assets (prompts, resources, tools) when scanning

ablox-chetan:
- Added a guarded McpError import plus _is_missing_capability_error so we can detect “method not found” style responses even when the MCP dependency is absent
  - Remote bulk scans for tools, prompts, and resources now short-circuit with empty results instead of propagating server errors when a capability is missing
  - Remote single-item scans raise a clear ValueError when the server omits the requested feature, preventing 500s while still informing the caller
  - Stdio scan paths mirror the same missing-capability handling so CLI usage behaves consistently


harishsg999 (approved):
lgtm

PR #34, opened 2025-10-22
adding bedrock support

nishadeborahphilips:
This PR adds support for AWS Bedrock LLM models with multiple authentication options while maintaining full backward compatibility with existing OpenAI and Anthropic providers.

Key Features

  - Multiple Authentication Methods: Supports Bedrock API keys, AWS profiles, IAM roles, and temporary session tokens
  - Smart Authentication Detection: Automatically detects Bedrock models (via bedrock/ prefix) and applies appropriate authentication strategy
  - Environment Variable Fallbacks: AWS region, profile, and credentials automatically resolve from standard AWS environment variables
  - Enhanced Error Handling: Dedicated handling for Bedrock-specific exceptions (AccessDenied, ThrottlingException, etc.)

  Authentication Options

  1. Bedrock API Key: MCP_SCANNER_LLM_API_KEY
  2. AWS Profile: AWS_PROFILE environment variable
  3. IAM Role/Temporary Credentials: Standard AWS credential chain

  Files Changed

  - mcpscanner/config/config.py: Added AWS configuration parameters
  - mcpscanner/config/constants.py: Added AWS environment variable constants
  - mcpscanner/core/analyzers/llm_analyzer.py: Implemented Bedrock authentication and error handling
  - mcpscanner/core/scanner.py: Updated analyzer initialization for Bedrock support

Integration tests:
Screenshot 2025-10-25 at 6 42 31 PM
Screenshot 2025-10-25 at 6 42 52 PM

### DIFF
--- a/mcpscanner/core/analyzers/llm_analyzer.py
+++ b/mcpscanner/core/analyzers/llm_analyzer.py
@@ -105,6 +105,7 @@ class LLMAnalyzer(BaseAnalyzer):
         self._aws_region = config.aws_region_name if is_bedrock else None
         self._aws_session_token = config.aws_session_token if is_bedrock else None
         self._aws_profile_name = config.aws_profile_name if is_bedrock else None
+        self._llm_timeout = config.llm_timeout
 
         # Load shared protection rules and analysis prompts
         self._protection_rules = self._load_prompt(
@@ -493,7 +494,7 @@ class LLMAnalyzer(BaseAnalyzer):
                     "messages": messages,
                     "max_tokens": self._max_tokens,
                     "temperature": self._temperature,
-                    "timeout": 30.0,
+                    "timeout": self._llm_timeout,
                 }
 
                 # Add API key if set (works for OpenAI, Anthropic, and Bedrock API keys)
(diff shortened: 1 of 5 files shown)


mcpscanner/core/analyzers/llm_analyzer.py
```diff
@@ -484,13 +510,51 @@ async def _make_llm_request(
...
                 last_exception = e
                 error_msg = str(e).lower()
 
+                # Check for AWS/Bedrock specific errors (only for Bedrock models)
+                is_bedrock_model = self._model and "bedrock/" in self._model
+                if is_bedrock_model and any(
+                    keyword in error_msg
+                    for keyword in [
+                        "bedrockexception",
+                        "accessdenied",
```
asamba-cisco: I'd suggest retrying only for recoverable errors like throttling / rate limits. In general, Access denied and unauthorized operation are hard failure and not worth retrying unless its applicable for Bedrock for some reason.

rucpande (changes requested):
Thanks for adding support for auth methods for AWS in addition to the API key. 

Do you think it makes sense to also support any of these auth methods in the REST API mode? That will still fail if the MCP_SCANNER_LLM_API_KEY is not present. Take a look at mcpscanner/api/api.py please.

Also could you add integration tests to test these new auth modes, run 'pytest -m integration' and add screenshots please. tests/test_openai_llm_integration.py has an example for OpenAI.

rucpande (approved):

PR #47, opened 2025-10-25
Performance: add tracing, concurrency controls, HTTP reuse, and caching

vsubra1:
**Summary**
This PR introduces optional lightweight tracing, configurable concurrency, and several low-risk optimizations that improve scan throughput and stability, particularly for prompts/resources. It also adds connection reuse for API calls and caches YARA compilation and Scanner instances in API mode.
Reason for this change : It was noticed that sometimes the MCP Scanner took more than 60 seconds to scan a MCP Serverw with large number of tools/prompts/resources. This will help in reducing this time

**Key changes**
**Tracing**
Add JSON span tracing utility (stdout) and --trace flag
Trace hot paths in scanner, analyzers, API routes

Tracing was added to measure the effect of the perf improvements. It will be helpful for users to see where the time is spent in the scan if needed

**Performance Improvements**

- Parallelize prompts/resources with a bounded semaphore (default 8)
- Bound tools concurrency to prevent provider spikes

Add env/CLI overrides:
MCP_SCANNER_MAX_CONCURRENCY_TOOLS / --max-concurrency-tools
MCP_SCANNER_MAX_CONCURRENCY_PROMPTS / --max-concurrency-prompts
MCP_SCANNER_MAX_CONCURRENCY_RESOURCES / --max-concurrency-resources

- HTTP reuse
- Persistent httpx.AsyncClient(http2=True) in ApiAnalyzer for connection pooling
- Caching/stability
- Cache compiled YARA rules at module level keyed by rules dir + mtimes
- Cache Scanner instances in API factory (per rules_path, api_on, llm_on)
- Micro-optimizations
- Avoid JSON roundtrips (use model_dump with fallback)
- List+join for resource text aggregation

**Docs**
README: tracing and concurrency usage
Development guide already documents tracing

**Performance results (local complete server)**
Tools (6): total ~398 ms → ~340 ms (≈14–15% faster)
Prompts (5): total ~1.0 s → ~0.22 s (≈4–5× faster) due to parallelization
Resources (5): total ~1.0 s → ~0.29 s (≈3–4× faster) due to parallelization
Backward compatibility
Tracing is opt-in; disabled by default
Concurrency defaults remain modest (8)
API behavior unchanged; performance improved via caching

**Tests**
185/186 tests pass without external keys
The remaining test (OpenAI integration) requires a valid OPENAI_API_KEY
To run all: export OPENAI_API_KEY=sk-...; pytest -q
Or skip: pytest -q -k 'not test_openai_llm_analyzer_detects_risk'

**Changelog**
Added: Optional JSON tracing (MCP_SCANNER_TRACING=1, --trace)
Added: Configurable concurrency for tools/prompts/resources (env + CLI flags)
Improved: Prompts/resources scanning parallelized with bounded concurrency
Improved: Persistent HTTP client for API analyzer (HTTP/2 pooling)
Improved: YARA rules caching across analyzers
Improved: Scanner caching in API factory to reduce per-request overhead
Improved: Avoided JSON roundtrips for MCP models; faster resource text joins
Docs: Updated README with tracing and concurrency options

### DIFF
--- a/mcpscanner/cli.py
+++ b/mcpscanner/cli.py
@@ -774,11 +774,11 @@ async def main():
             cfg = _build_config(selected_analyzers)
             # Override concurrency if provided
             if args.max_concurrency_tools is not None:
-                cfg._max_concurrency_tools = args.max_concurrency_tools
+                cfg.max_concurrency_tools = args.max_concurrency_tools
             if args.max_concurrency_prompts is not None:
-                cfg._max_concurrency_prompts = args.max_concurrency_prompts
+                cfg.max_concurrency_prompts = args.max_concurrency_prompts
             if args.max_concurrency_resources is not None:
-                cfg._max_concurrency_resources = args.max_concurrency_resources
+                cfg.max_concurrency_resources = args.max_concurrency_resources
             scanner = Scanner(cfg, rules_dir=args.rules_path)
             auth = Auth.bearer(args.bearer_token) if args.bearer_token else None
             results_raw = await scanner.scan_remote_server_tools(
@@ -789,11 +789,11 @@ async def main():
         elif args.cmd == "stdio":
             cfg = _build_config(selected_analyzers)
             if args.max_concurrency_tools is not None:
-                cfg._max_concurrency_tools = args.max_concurrency_tools
+                cfg.max_concurrency_tools = args.max_concurrency_tools
             if args.max_concurrency_prompts is not None:
-                cfg._max_concurrency_prompts = args.max_concurrency_prompts
+                cfg.max_concurrency_prompts = args.max_concurrency_prompts
             if args.max_concurrency_resources is not None:
-                cfg._max_concurrency_resources = args.max_concurrency_resources
+                cfg.max_concurrency_resources = args.max_concurrency_resources
             scanner = Scanner(cfg, rules_dir=args.rules_path)
             env_dict = {}
             for item in args.stdio_env or []:
@@ -821,11 +821,11 @@ async def main():
         elif args.cmd == "config":
             cfg = _build_config(selected_analyzers)
             if args.max_concurrency_tools is not None:
-                cfg._max_concurrency_tools = args.max_concurrency_tools
+                cfg.max_concurrency_tools = args.max_concurrency_tools
             if args.max_concurrency_prompts is not None:
-                cfg._max_concurrency_prompts = args.max_concurrency_prompts
+                cfg.max_concurrency_prompts = args.max_concurrency_prompts
             if args.max_concurrency_resources is not None:
-                cfg._max_concurrency_resources = args.max_concurrency_resources
+                cfg.max_concurrency_resources = args.max_concurrency_resources
             scanner = Scanner(cfg, rules_dir=args.rules_path)
             auth = Auth.bearer(args.bearer_token) if args.bearer_token else None
             scan_results = await scanner.scan_mcp_config_file(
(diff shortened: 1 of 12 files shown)


mcpscanner/core/scanner.py
```diff
@@ -1753,102 +1780,108 @@ async def scan_remote_server_resources(
...
                             analyzers=[],
                             findings=[],
-                        ))
-                        continue
+                        )
 
                     # Analyze the resource
                     try:
-                        result = await self._analyze_resource(
+                        async with get_tracer().span("scanner.analyze.resource", {"uri": str(resource.uri)}):
```
rucpande: don't we need to wrap this in a semaphore?

mcpscanner/core/scanner.py
rucpande: worth using max_concurrency here as well?

rucpande (commented):
nice changes. haven't reviewed everything, but added few quick comments.

mcpscanner/core/scanner.py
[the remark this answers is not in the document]
vsubra1: _clamp_concurrency() static method that enforces [1, 8] range
Applied to all concurrency settings: tools, prompts, resources, and LLM
CLI overrides now use property setters that automatically clamp values

mcpscanner/config/config.py
```diff
@@ -95,6 +100,23 @@ def __init__(
         self._oauth_client_secret = oauth_client_secret
         self._oauth_token_url = oauth_token_url
         self._oauth_scopes = oauth_scopes
+        self._tracing_enabled = tracing_enabled
+        # Concurrency (allow env overrides if not explicitly provided)
+        self._max_concurrency_tools = (
```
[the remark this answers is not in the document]
vsubra1: Resource analysis is now wrapped in semaphore at line 1974:
async with sem:    async with get_tracer().span("scanner.analyze.resource", ...):        result = await self._analyze_resource(...)

mcpscanner/core/scanner.py
vsubra1: Fixed! Resource analysis is now wrapped in a semaphore to respect max_concurrency_resources bounds.

The code at line 1988 now wraps _analyze_resource() in the semaphore:
```python
# Analyze the resource (bounded by semaphore)
try:
    async with sem:
        async with get_tracer().span("scanner.analyze.resource", {"uri": str(resource.uri)}):
            result = await self._analyze_resource(
                text_content,
                resource.uri,
                resource.name or "",
                resource.description or "",
                resource.mimeType or "unknown",
                analyzers,
                http_headers,
            )
        return result
```

This ensures both resource reading and analysis respect the concurrency limit.

mcpscanner/utils/tracing.py
```diff
@@ -0,0 +1,155 @@
+"""
```
sanket-mendapara: Can you add a Cisco license comment blob in the newly added files, like all other scanner files?

tests/test_httpx_logger_sync.py
```diff
@@ -0,0 +1,65 @@
+import asyncio
```
sanket-mendapara: Can you add a Cisco license comment blob in the newly added files, like all other scanner files?

mcpscanner/api/api.py
```diff
@@ -129,13 +136,21 @@ def scanner_factory(
             HTTPException: If API scan is requested but config is invalid.
         """
         api_key, endpoint_url, llm_api_key = _prepare_scanner_config(analyzers)
+        # Cache key based on rules path and which analyzers require keys
+        key = (rules_path or "", bool(api_key), bool(llm_api_key))
```
rucpande: Based on the Bedrock PR and another one in the works for local LLMs, the llm api key will be optional for the LLM analyzer.  We might need another value here?

mcpscanner/cli.py
```diff
@@ -863,6 +922,12 @@ async def main():
 
         elif args.cmd == "resources":
             cfg = _build_config(selected_analyzers)
+            if args.max_concurrency_tools is not None:
+                cfg._max_concurrency_tools = args.max_concurrency_tools
```
rucpande: For this and the next 2 assignments - it should be cfg.max_concurrency_*.   not the private _max_concurrency_* field to use the new clamping logic.

mcpscanner/core/scanner.py
```diff
@@ -846,6 +886,121 @@ async def scan_remote_server_tools(
...
+                        resource_uri=r.uri,
+                        resource_name=r.name or "",
+                        resource_mime_type=r.mimeType or "unknown",
+                        status="skipped",
+                        analyzers=[],
+                        findings=[],
+                    )
+                async with res_sem:
+                    async with get_tracer().span("scanner.analyze.resource", {"uri": str(r.uri)}):
+                        return await self._analyze_resource(
```
rucpande: try..catch

asamba-cisco:
Inactive.

PR #85, opened 2025-12-17
refact: Updating constraints with litellm and python versions & bumping release version

sanket-mendapara:
- Updating litellm version bound to >=1.77.0
- Removing upper bound for python versioning.
- Bumping release version to 3.2.3


harishsg999 (approved):
lgtm

PR #98, opened 2026-01-11
fix: escape percent signs in --expand-vars help string

barryqy:
## Problem
`mcp-scanner --help` fails with `ValueError: unsupported format character 'V' (0x56) at index 142` on all Python versions (3.11-3.14).

## Root Cause
The `--expand-vars` argument help string (line 998) contains `%VAR%` as an example of Windows environment variable syntax. argparse interprets `%` in help strings as format specifiers unless escaped, causing a ValueError.

## Solution
Escaped the percent signs: `%VAR%` → `%%VAR%%`

In Python string formatting, `%%` escapes to a single `%`, so the help text correctly displays `%VAR%` to users.

## Changes
- Changed line 998 in `mcpscanner/cli.py`
- One character change: `%VAR%` → `%%VAR%%`

## Testing
- ✅ `mcp-scanner --help` now works without ValueError
- ✅ Help text displays correctly
- ✅ Works on Python 3.11, 3.12, 3.13, 3.14

## Impact
Fixes the issue where users report "it used to work before". This bug was introduced when `--expand-vars` was added after commit `164d5ea`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,7 +16,8 @@ The SDK is designed to be easy to use while providing powerful scanning capabili
 - **Multiple Modes:** Run scanner as a stand-alone CLI tool or REST API server
 - **Multi-Engine Security Analysis**: Use all three scanning engines together or independently based on your needs.
 - **Comprehensive Scanning**: Scan MCP tools, prompts, resources, and server instructions for security findings
-- **Behavioural Code Scannig**: Scan Source code of MCP servers for finding threats.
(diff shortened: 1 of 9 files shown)


sanjayab23 (approved):

harishsg999 (approved):
lgtm

harishsg999:
hey @barryqy thanks for PR 
can you  please try rebase with main again and then push to fix this CodeQL issue
Thanks!

PR #103, opened 2026-01-15
Add Docker support for CLI and API with multi-arch builds

mhajder:
Introduce Dockerfiles for both CLI and API deployments, enhancing security and reproducibility. Implement a CI workflow for automated multi-architecture image builds and publishing to the GitHub Container Registry. Update dependencies for arm64 support.


ablox-chetan:
I reviewed this PR with extra scrutiny because it adds a new publishing workflow and container build path. I do not think it should be merged as-is.

1. `GHCR image name omits the required namespace`
The workflow currently sets `images:` to values that expand to `ghcr.io/mcp-scanner` and `ghcr.io/mcp-scanner-api`. For GitHub Container Registry, images should be scoped as `ghcr.io/<owner>/<image>`.

Suggested fix:
- Keep `IMAGE_NAME` and `IMAGE_NAME_API` as bare names if you want, but build the final image reference as `${{ env.REGISTRY }}/${{ github.repository_owner }}/${{ env.IMAGE_NAME }}` and `${{ env.REGISTRY }}/${{ github.repository_owner }}/${{ env.IMAGE_NAME_API }}`.
- Alternatively, hardcode the owner/org namespace if that is preferred.

2. `Privileged workflow should pin actions by full SHA`
This job has `packages: write` and invokes several third-party actions using mutable version tags. For a workflow that can publish artifacts, that is a supply-chain risk.

Suggested fix:
- Pin every `uses:` reference in this workflow to a full 40-character commit SHA, including `actions/checkout`, `docker/setup-qemu-action`, `docker/setup-buildx-action`, `docker/login-action`, `docker/metadata-action`, and `docker/build-push-action`.
- If desired, use Dependabot or Renovate to keep those pins updated safely.

3. `New container publisher bypasses release gating`
This adds a second artifact publishing path that runs on every push to `main`, but unlike the existing PyPI publish flow it does not use a protected environment or reviewer gate. For an untrusted-author PR, merging becomes the point at which the new workflow and Dockerfiles acquire publish privileges.

Suggested fix:
- Add an `environment:` to the job and protect it with required reviewers, similar to the existing release path.
- Or restrict publishing to trusted release tags only, and keep branch pushes build-only.
- Or split this into two workflows: one that builds/tests on branch activity and one gated workflow that publishes.

Given that this PR also includes substantial `uv.lock` churn, I would strongly prefer the workflow hardening and publish-path controls to be resolved before merge, and ideally keep dependency refreshes separate from CI publishing changes.

mhajder:
Hi @ablox-chetan

1. Fixed, thanks for noticing.
2. Added full SHA for all actions in this workflow. In the future, you can use Dependabot or Renovate. I can prepare a PR, but you will have to decide which you prefer to use.
3. Added `environment: release` and also limited docker build only for tags.

I changed the base docker image from `python:3.14-alpine3.23` to `python:3.13-slim` so that it builds without problems also for arm64. 

The changes were force-pushed to the same branch.

PR #104, opened 2026-01-15
Update pre-commit hooks and format code

mhajder:
Update pre-commit hook versions for better compatibility with Python and ensure code formatting across all files.

This PR is just:
```sh
pre-commit autoupdate # reverted only "addlicense" hook to the old used version because latest one breaks files
pre-commit run --all-files
```


harishsg999 (approved):
lgtm

PR #118, opened 2026-02-04
fix(cli): display correct server_label for behavioral command in detailed mode

dsonyy:
## Summary
- Fixed a display bug where the `--detailed` flag code path was missing the behavioral command check
- This caused the CLI to show the default server URL (`https://mcp.deepwiki.com/mcp`) instead of the actual source path when running `mcp-scanner behavioral <path> --detailed`

## Problem
When running:
```bash
python -m mcpscanner.cli behavioral /path/to/mcp/server --detailed
```

The output incorrectly showed:
```
Server URL: https://mcp.deepwiki.com/mcp
```

Instead of:
```
Server URL: behavioral:/path/to/mcp/server
```

## Root Cause
The non-detailed code path (lines 1942-1943) had the behavioral check:
```python
elif hasattr(args, "cmd") and args.cmd == "behavioral":
    server_label = f"behavioral:{args.source_path}"
```

But the detailed code path (starting at line 2041) was missing this check.

## Fix
Added the missing behavioral command check to the detailed code path.

## Test plan
- [x] Run `mcp-scanner behavioral <path> --detailed` and verify server_label shows correctly
- [x] Run `mcp-scanner behavioral <path> --verbose` and verify it still works

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,28 +43,50 @@ The SDK is designed to be easy to use while providing powerful scanning capabili
 - A valid Cisco AI Defense API Key (optional)
 - LLM Provider API Key (optional)
 
-### Installing from PyPI
+### Installing as a CLI tool
 
 ```bash
-uv venv -p <Python version less than or equal to 3.13> /path/to/your/choice/of/venv/directory
-source /path/to/your/choice/of/venv/directory/bin/activate
-uv pip install cisco-ai-mcp-scanner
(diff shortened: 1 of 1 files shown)


harishsg999 (approved):
lgtm 
Thanks @dsonyy for fix

PR #135, opened 2026-03-20
fix: tighten LLM threat analysis prompt to make it consistent 

harishsg999:
fix: tighten LLM threat analysis prompt to reduce non determinism in LLM results


sanket-mendapara (approved):
Detection numbers looks okay.

PR #136, opened 2026-03-24
feat: Add vulnerable packages analyzer for Python dependency scanning

harishsg999:
Integrate pip-audit as a new analyzer (VulnerablePackagesAnalyzer) that scans Python dependencies for known vulnerabilities (CVE/PYSEC/GHSA). The analyzer runs pip-audit as a subprocess, parses JSON output, and maps findings to the Cisco AI Threat Security Taxonomy (AITech-9.2 Supply Chain Compromise).

Key features:
- Scans requirements files, project directories, and installed environments
- Auto-discovers pip-audit binary with fallbacks: venv -> PATH -> uvx -> uv tool run
- Full uv package manager compatibility (uvx and uv tool run)
- CLI subcommand: mcp-scanner vulnerable-packages <path>
- Supports PyPI and OSV vulnerability services, auto-fix mode
- 53 unit tests covering discovery, parsing, taxonomy, and uv fallbacks
- Complete documentation with dedicated docs page


mcpscanner/core/analyzers/vulnerable_packages_analyzer.py
```diff
@@ -0,0 +1,315 @@
...
+
+    def analyze_requirements(self, requirements_path: str) -> List[SecurityFinding]:
+        """Scan a requirements file for vulnerable dependencies."""
+        if not self.enabled:
+            return []
+        path = Path(requirements_path)
+        if not path.is_file():
+            logger.warning("Requirements file does not exist: %s", requirements_path)
+            return []
+        args = ["-r", str(path), "--no-deps", "--disable-pip"]
```
sanket-mendapara: Why is `--no-deps` & `--disable-pip` is added here and why on just `pip-audit -r <requirements> by default` and only enabling `--no-deps/--disable-pip` behind an explicit opt-in for fully resolved inputs or pinned ones.

Here, currently every requirements-file scan is forced through `pip-audit -r ... --no-deps --disable-pip`. Per pip-audit’s own CLI docs, `--no-deps` skips dependency resolution and is only appropriate when the input is already fully resolved and pinned. On normal requirements.txt files this will miss vulnerable transitive packages and can produce false-clean results.
i.e. if `flask` is being checked then there can be deps like for ex. - `werkzeug`, `jinja2`, `urllib3` which might be installed as deps.
Here, it will not include/check transitive dependencies like above. I know it is a slow if we also check deps but atleast provide the option for users to run either full check or partial checks.

mcpscanner/core/analyzers/vulnerable_package_analyzer.py
```diff
@@ -0,0 +1,315 @@
...
+
+        if p.is_file() and p.suffix in (".txt", ".in"):
+            return self.analyze_requirements(str(p))
+
+        if p.is_dir():
+            req_file = p / "requirements.txt"
+            if req_file.is_file():
+                return self.analyze_requirements(str(req_file))
+
+            pyproject = p / "pyproject.toml"
```
sanket-mendapara, line 153: Here, project directories with pyproject.toml are scanned via `--path`, but `pip-audit --path` means **restrict to an installation path**, not **audit this project**. For a source repo without a checked-in `.venv`, this will audit the wrong target or nothing useful at all. pip-audit’s project mode is positional (`pip-audit .`) and lockfile mode is `pip-audit --locked ..`
-> One fix can be -
- for project directories, call `pip-audit <project_path>` or `pip-audit --locked <project_path>; reserve --path` for real installed environments only.

mcpscanner/cli.py
```diff
@@ -1975,11 +2121,19 @@ async def main():
...
-        results_dict = {
-            "server_url": server_label,
-            "scan_results": results,
-            "requested_analyzers": selected_analyzers,
-        }
+        is_vuln_pkg_scan = hasattr(args, "cmd") and args.cmd == "vulnerable-packages"
+        if is_vuln_pkg_scan:
+            results_dict = {
+                "mcp_server_repository": server_label,
+                "scan_results": results,
```
sanket-mendapara, line 2309: Below in results part, `vulnerable-packages --detailed` falls through the legacy display_results() path, which expects tool_name and server_url. This branch produces package_name and a synthetic repository label instead, so detailed output will show the default remote server URL and print package entries as Unknown.

Here, this can be handled by either special-case vulnerable-packages in the detailed branch the same way it is already special-cased in the summary branch, or route detailed output through ReportGenerator so package_name/mcp_server_repository are handled consistently.

Also, why is this results processing blob is handled here and not in ReportGenerator?

mcpscanner/core/report_generator.py
```diff
@@ -651,7 +651,7 @@ def _format_by_severity(
 
                 severity_groups[severity].append(
                     {
-                        "tool_name": result.get("tool_name", "Unknown"),
+                        "tool_name": result.get("tool_name", result.get("package_name", "Unknown")),
```
sanket-mendapara, line 665: Below in table output is still hardcoded to API/YARA/LLM or behavioral layouts. For `vulnerable-packages, --format table` will render irrelevant analyzer columns and no dedicated vulnerable-packages column, which is misleading for users and inconsistent with the new command semantics.
Maybe add a vulnerable-packages-specific table layout, for ex - Package, Status, VULNERABLE_PACKAGES, Severity, and use that whenever vulnerable_packages_analyzer is present.

mcpscanner/core/report_generator.py
```diff
@@ -186,7 +186,7 @@ def __init__(self, scan_data: Union[Dict[str, Any], str]):
         else:
             self.data = scan_data
 
-        self.server_url = self.data.get("server_url", "Unknown")
+        self.server_url = self.data.get("server_url", self.data.get("mcp_server_repository", "Unknown"))
         self.scan_results = self.data.get("scan_results", [])
         self.requested_analyzers = self.data.get("requested_analyzers", [])
```
sanket-mendapara, line 196: This new analyzer was never added to `requested_analyzer_keys` or `analyzer_stats`. As a result, `--stats` omits `vulnerable_packages_analyzer` entirely, and any analyzer-aware formatting remains partially unaware of the new analyzer.
Register VULNERABLE_PACKAGES in requested_analyzer_keys, and add vulnerable_packages_analyzer to analyzer_stats so reporting stays accurate across all formats.

docs/vulnerable-packages-scanning.md
```diff
@@ -0,0 +1,236 @@
...
+
+## How It Works
+
+### Scan Targets
+
+The analyzer uses the following heuristics to determine how to scan a path:
+
+1. **File ending in `.txt` or `.in`**: Treated as a requirements file (`pip-audit -r <path>`)
+2. **Directory with `requirements.txt`**: Scans the requirements file
+3. **Directory with `pyproject.toml`**: Scans the project environment. If a `.venv` directory exists, uses `--path .venv`; otherwise uses `--path <directory>`
```
sanket-mendapara: This currently documents the same incorrect `--path` behavior as the implementation. Once the scan mode is fixed, this doc should be updated to distinguish project scans from installed-environment scans.

mcpscanner/cli.py
```diff
@@ -1975,11 +2121,19 @@ async def main():
...
 
-        results_dict = {
-            "server_url": server_label,
-            "scan_results": results,
-            "requested_analyzers": selected_analyzers,
-        }
+        is_vuln_pkg_scan = hasattr(args, "cmd") and args.cmd == "vulnerable-packages"
+        if is_vuln_pkg_scan:
+            results_dict = {
+                "mcp_server_repository": server_label,
```
sanket-mendapara: This and the docs use mcp_server_repository for a value like vulnerable-packages:/path/to/requirements.txt, which is not actually a repository identifier. `scan_target` or `source_path` would be a clearer convention or something more relevantly descriptive of package scan artifacts

tests/test_vulnerable_packages_analyzer.py
```diff
@@ -0,0 +1,751 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# SPDX-License-Identifier: Apache-2.0
+
+"""Unit tests for the vulnerable packages analyzer."""
```
sanket-mendapara: This file currently covers analyzer internals, but not CLI/report integration. Adding focused tests for --detailed, --table, and --stats would likely catch the reporting regressions mentioned above in the PR.

mcpscanner/core/analyzers/vulnerable_package_analyzer.py
```diff
@@ -0,0 +1,315 @@
...
+        vuln: Dict[str, Any],
+        source: str,
+    ) -> SecurityFinding:
+        """Create a SecurityFinding for a single vulnerability."""
+        vuln_id = vuln.get("id", "UNKNOWN")
+        fix_versions = vuln.get("fix_versions", [])
+        aliases = vuln.get("aliases", [])
+        description = vuln.get("description", "")
+
+        has_fix = len(fix_versions) > 0
```
sanket-mendapara, line 330: Here, sets severity from “has a fix” rather than advisory severity. That may be intentional, but it should be documented explicitly; otherwise consumers will assume it reflects CVSS-like severity. Maybe using more descriptive naming conventions.

sanket-mendapara (changes requested):

docs/mcp-threats-taxonomy.md
```diff
@@ -59,6 +65,7 @@ The MCP Scanner maps all detected threats to the MCP Taxonomy, which provides a
...
 | AITech-12.1 | Tool Exploitation | AISubtech-12.1.2 | Tool Poisoning | Corrupting, modifying, or degrading the functionality, outputs, or behavior of tools used by agents through data poisoning, configuration tampering, or behavioral manipulation, causing the tool resulting in deceptive or malicious outputs, privilege escalation, or propagation of altered data. |
 | AITech-12.1 | Tool Exploitation | AISubtech-12.1.5 | Tool Shadowing | Disguising, substituting or duplicating legitimate tools within an agent or MCP server or tool registry, enabling malicious tools with identical or similar identifiers to intercept or replace trusted tool calls, leading to unauthorized actions, data exfiltration, or redirection of legitimate operations. |
+| AITech-9.2 | Detection Evasion | AISubtech-9.2.1 | Supply Chain Compromise | A Python dependency with a publicly known vulnerability (CVE/PYSEC/GHSA) was detected. Vulnerable dependencies in MCP server packages can be exploited to compromise the server, exfiltrate data, or escalate privileges. |
```
sanket-mendapara: This is a mismatch. Check the taxonomy category again and fix this.

docs/api-reference.md
```diff
@@ -28,7 +28,7 @@ The MCP Scanner API supports all the same output formatting options as the CLI t
 | Parameter | Type | Default | Description |
 |-----------|------|---------|-------------|
 | `server_url` | string | required | URL of the MCP server to scan |
-| `analyzers` | array | ["yara"] | List of analyzers to use: "api", "yara", "llm" |
+| `analyzers` | array | ["yara"] | List of analyzers to use: "api", "yara", "llm", "vulnerable_packages" |
```
sanket-mendapara: Here, we can shorten the package name like vulnerable_pkg or something similar. Thoughts!?
Also, for all the reference use singular then plural like package instead of packages.

sanket-mendapara (changes requested):

sanket-mendapara (approved):

PR #137, opened 2026-03-24
added TUI for MCP Scanner 

harishsg999:
added TUI for MCP Scanner using textual

Screenshot 2026-03-23 at 6 51 34 PM


mcpscanner/tui/screens/results.py
```diff
@@ -0,0 +1,212 @@
...
+                    yield Button("New Scan", id="new-scan-btn", variant="success")
+
+            yield StatsBar(self.scan_results)
+            yield DataTable(id="results-table", cursor_type="row", zebra_stripes=True)
+            yield Static("Select a row to view details.", id="detail-panel")
+        yield Footer()
+
+    def on_mount(self) -> None:
+        table = self.query_one("#results-table", DataTable)
+        table.add_columns("Status", "Name", "Severity", "Analyzer", "Threat")
```
sanket-mendapara: Instead of just `Name` probably have `Tool Name`.

mcpscanner/tui/widgets/stats_bar.py
```diff
@@ -0,0 +1,52 @@
...
+        unsafe = total - safe
+
+        severity_counts: dict[str, int] = {}
+        for r in self.results:
+            for analyzer_data in r.get("findings", {}).values():
+                sev = analyzer_data.get("severity", "SAFE")
+                if sev != "SAFE":
+                    severity_counts[sev] = severity_counts.get(sev, 0) + 1
+
+        high = severity_counts.get("HIGH", 0) + severity_counts.get("UNKNOWN", 0)
```
sanket-mendapara: severity_counts.get("HIGH", 0) + severity_counts.get("UNKNOWN", 0) why this both count is added to high? Unknown should have different count!?

mcpscanner/tui/widgets/results_table.py
```diff
@@ -0,0 +1,83 @@
...
+
+            table.add_row(status, name, severity_display, analyzer_display, threat_display)
+
+    def on_data_table_row_highlighted(self, event: DataTable.RowHighlighted) -> None:
+        if event.cursor_row is not None and 0 <= event.cursor_row < len(self.results):
+            self.post_message(self.RowSelected(self.results[event.cursor_row]))
+
+
+def _highest_severity(severities: list[str]) -> str:
+    order = {"HIGH": 0, "UNKNOWN": 0, "MEDIUM": 1, "LOW": 2, "SAFE": 3}
```
sanket-mendapara: why are we considering high and unknown as same level?
"HIGH": 0, "UNKNOWN": 0

sanket-mendapara (changes requested):
Why do we need status & severity both? Can we have status at top along with the server details where we show that server is safe or unsafe? Also, we can have an AITECH ID - Name in column.
Also, at the top of the TUI it says results which has server URL.
Can we change it to Server Artifact and show URL in case we are scanning remote server and show filename or path in case we are scanning local file or config?

mcpscanner/core/report_generator.py
```diff
@@ -793,15 +810,18 @@ def get_analyzer_status(analyzer_key):
...
             }
 
             if findings:
-                severities = [f.get("severity", "SAFE") for f in findings.values()]
-                severity_text = self._get_highest_severity(severities)
+                severities = [
+                    f.get("severity", "UNKNOWN") for f in findings.values()
+                ]
+                severity_text = get_highest_severity(severities)
                 severity_emoji = severity_emojis.get(severity_text, "🟢")
```
sanket-mendapara: Have default emoji as unknown colored and not safe if default is not known then.

sanket-mendapara (changes requested):

PR #141, opened 2026-03-26
feat: add Docker-sandboxed PyPI package scanner with behavioral analysis

harishsg999:
Summary
Adds PyPI and npm package scanning (Docker-sandboxed by default, optional local SDK mode) and tightens meta-analyzer false-positive filtering for benign publisher usage documentation.

Package scanning (PyPI + npm)
New CLI subcommands: pypi-scan and npm-scan, with --no-docker, --rebuild-image, --detailed, and the same output/format flags as other analyzers
SDK entrypoints: PyPIPackageScanner and NPMPackageScanner with sync (scan_package) and async (scan_package_async) APIs
Docker mode (default): isolated containers download archives, extract safely, and run behavioral analysis; results returned as JSON
Local/SDK mode: HTTPS-only download, host allow-listing, archive/extraction caps, tarfile data filter; package code is never executed (Python parsed, JS/TS parsed via tree-sitter)
Shared safety primitives in package_sandbox.py: PEP 508 / npm name validation, SSRF defense-in-depth (private/link-local/unspecified DNS checks), digest verification
Docker hardening: digest-pinned base image, non-root user, capability drops, read-only root + tmpfs, versioned image tags scoped per ecosystem (MCP_SCANNER_DOCKER_IMAGE_TAG vs MCP_SCANNER_NPM_DOCKER_IMAGE_TAG)
Fail-closed behavior: missing LLM key, degraded scans, and unreliable analyzer stats surface scan_status: "error" and is_safe: null instead of false-safe results
SAFE behavioral placeholders excluded from total_findings / is_safe counts
CLI --no-docker path uses scan_package_async() inside the async CLI main loop
Meta-analyzer
Updates meta_analysis_prompt.md to filter LLM prompt-injection false positives on first-party publisher usage docs (e.g. "When to Use", "Always use this tool for …") when text is legitimate workflow guidance
Keeps real threats: jailbreak/override language, context-harvesting/exfiltration, and corroborated findings from other analyzers
Contract tests pin the prompt safeguards
Docs
docs/pypi-scanning.md — PyPI scanning CLI/SDK, config, security, output schema, error codes
docs/npm-scanning.md — npm scanning CLI/SDK, config, security, output schema, error codes
README.md — overview and quick-start for both commands
Test plan
 tests/test_pypi_scanner.py — Docker/local paths, async entrypoint, SAFE finding counts, host allow-list, degraded scans, image tag scoping
 tests/test_npm_scanner.py — parity with PyPI scanner, registry host checks, local/docker modes
 tests/test_meta_analyzer.py — publisher-usage FP prompt contract
 Manual: mcp-scanner pypi-scan flask (Docker)
 Manual: mcp-scanner npm-scan @modelcontextprotocol/server-everything (Docker)
 Manual: mcp-scanner pypi-scan flask --no-docker with MCP_SCANNER_LLM_API_KEY set
 Manual: re-scan Microsoft Learn MCP server and confirm publisher-usage docs no longer flip tools unsafe via meta-analysis

## Summary by CodeRabbit

- **New Features**
  - Added PyPI and npm package scanning via CLI (`pypi-scan`, `npm-scan`) and SDK, with Docker-isolated mode by default and an optional local/no-Docker mode.
  - Supports scanning specific versions, saving JSON results, and configurable output/verbosity.
  - Added JavaScript/TypeScript behavioral analysis for extracted package sources (no package execution).

- **Bug Fixes**
  - Improved fail-closed safety reporting: degraded/error scans are no longer marked safe.

- **Documentation**
  - Updated README and docs with new PyPI/npm scanning guides, examples, and mode/security details, plus false-positive clarification.

- **Tests**
  - Added new test coverage for PyPI scanner, npm scanner, and JavaScript context extraction.

### DIFF
--- a/mcpscanner/cli.py
+++ b/mcpscanner/cli.py
@@ -1073,6 +1073,39 @@ async def main():
         help="Output format (default: %(default)s)",
     )
 
+    # PyPI package scan subcommand (Docker-sandboxed)
+    p_pypi = subparsers.add_parser(
+        "pypi-scan",
+        help="Download and scan a PyPI package in a Docker sandbox",
+    )
+    p_pypi.add_argument("package", help="PyPI package name (e.g., flask)")
+    p_pypi.add_argument(
+        "--version", help="Specific package version (default: latest)"
+    )
+    p_pypi.add_argument(
+        "--output", "-o", help="Save scan results to a file"
+    )
+    p_pypi.add_argument(
+        "--verbose", "-v", action="store_true", help="Print verbose output"
+    )
+    p_pypi.add_argument(
+        "--raw", "-r", action="store_true", help="Print raw JSON output"
+    )
+    p_pypi.add_argument(
+        "--format",
+        choices=[
+            "raw", "summary", "detailed", "by_tool",
+            "by_analyzer", "by_severity", "table",
+        ],
+        default="summary",
+        help="Output format (default: %(default)s)",
+    )
+    p_pypi.add_argument(
+        "--rebuild-image",
+        action="store_true",
+        help="Force rebuild of the Docker scanner image",
+    )
+
     # Stdio subcommand
     p_stdio = subparsers.add_parser(
         "stdio", help="Scan an MCP server via stdio (local command execution)"
@@ -1951,6 +1984,69 @@ async def main():
                 if args.verbose:
                     print(f"Results saved to {args.output}")
 
+        elif args.cmd == "pypi-scan":
+            from mcpscanner.core.pypi_scanner import (
+                DockerNotAvailableError,
+                PyPIPackageScanner,
+                PyPIScanError,
+            )
+
+            try:
+                scanner = PyPIPackageScanner()
+                if getattr(args, "rebuild_image", False):
+                    scanner.build_image(force=True)
+
+                scan_results = scanner.scan_package(
+                    package=args.package,
+                    version=getattr(args, "version", None),
+                    verbose=getattr(args, "verbose", False),
+                )
+
+                pkg_spec = args.package
+                if getattr(args, "version", None):
+                    pkg_spec = f"{args.package}=={args.version}"
+
+                results = []
+                for finding in scan_results.get("findings", []):
+                    analyzer_name = finding.get("analyzer", "unknown") + "_analyzer"
+                    results.append({
+                        "tool_name": finding.get("details", {}).get("function_name", pkg_spec),
+                        "tool_description": finding.get("summary", ""),
+                        "status": "completed",
+                        "is_safe": False,
+                        "findings": {
+                            analyzer_name: {
+                                "severity": finding.get("severity", "UNKNOWN"),
+                                "threat_summary": finding.get("summary", ""),
+                                "threat_names": [finding.get("threat_category", "UNKNOWN")],
+                                "total_findings": 1,
+                                "mcp_taxonomies": [],
+                            }
+                        },
+                    })
+
+                if not results:
+                    results.append({
+                        "tool_name": pkg_spec,
+                        "tool_description": f"PyPI package scan of {pkg_spec}",
+                        "status": "completed",
+                        "is_safe": True,
+                        "findings": {},
+                    })
+
+                if args.output:
+                    with open(args.output, "w", encoding="utf-8") as f:
+                        json.dump(results, f, indent=2)
+                    if args.verbose:
+                        print(f"Results saved to {args.output}")
+
+            except DockerNotAvailableError as e:
+                print(f"Error: {e}", file=sys.stderr)
+                sys.exit(1)
+            except PyPIScanError as e:
+                print(f"Scan Error: {e}", file=sys.stderr)
+                sys.exit(1)
+
         # Backward compatibility path (no subcommand used)
         elif args.stdio_command:
             cfg = _build_config(selected_analyzers)
@@ -2085,6 +2181,11 @@ async def main():
             server_label = f"virustotal:{args.scan_path}"
         elif hasattr(args, "cmd") and args.cmd == "behavioral":
             server_label = f"behavioral:{args.source_path}"
+        elif hasattr(args, "cmd") and args.cmd == "pypi-scan":
+            pkg_spec = args.package
+            if getattr(args, "version", None):
+                pkg_spec = f"{args.package}=={args.version}"
+            server_label = f"pypi:{pkg_spec}"
         elif AnalyzerEnum.BEHAVIORAL in selected_analyzers and args.source_path:
             server_label = f"behavioral:{args.source_path}"
         elif args.stdio_command:
@@ -2198,6 +2299,11 @@ async def main():
             server_label = "well-known-configs"
         elif hasattr(args, "cmd") and args.cmd == "behavioral":
             server_label = f"behavioral:{args.source_path}"
+        elif hasattr(args, "cmd") and args.cmd == "pypi-scan":
+            pkg_spec = args.package
+            if getattr(args, "version", None):
+                pkg_spec = f"{args.package}=={args.version}"
+            server_label = f"pypi:{pkg_spec}"
 
         # Handle prompts, resources, and instructions with detailed view
         if hasattr(args, "cmd") and args.cmd == "prompts":
--- a/mcpscanner/cli.py
+++ b/mcpscanner/cli.py
@@ -120,6 +120,50 @@ def _create_auth_with_headers(
         return Auth.custom(custom_headers)
 
 
+def _package_scan_to_tool_results(
+    *,
+    scan_results: dict,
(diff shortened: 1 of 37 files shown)


mcpscanner/cli.py
```diff
@@ -119,6 +119,50 @@ def _create_auth_with_headers(
         return Auth.custom(custom_headers)
 
 
+def _package_scan_to_tool_results(
```
sanket-mendapara, line 122: `_package_scan_to_tool_results()` only looks at findings and defaults to safe when the list is empty. It never inspects scan_status or is_safe.
degraded input: {is_safe: None, scan_status: "error", findings: []}
CLI output    : [{... "status": "completed", "is_safe": True, "findings": {}}].

This path is reachable in both modes: the Docker success-path results in a degraded result without an "error" key (entrypoint.py:184-196), so _scan_in_docker's error check at pypi_scanner.py:352 doesn't fire and it returns the dict as-is; local mode returns the degraded dict directly. Net effect: a package whose LLM analysis failed for every function is shown to the user as a clean "completed / safe" scan. _package_scan_to_tool_results should propagate scan_status/is_safe (emit an "unknown/error" row and a non-zero exit code rather than a SAFE row).

mcpscanner/core/pypi_scanner.py
```diff
@@ -0,0 +1,670 @@
...
+            force: Rebuild even if the image already exists.
+        """
+        if not force and self._image_exists():
+            logger.info("Docker image %s already exists, skipping build", self._full_image)
+            return
+
+        logger.info("Building Docker image %s ...", self._full_image)
+
+        docker_dir = resources.files("mcpscanner.docker")
+        with resources.as_file(docker_dir) as ctx_path:
```
sanket-mendapara: `build_image()` builds from files it expects in the installed package. But package-data only ships prompts + yara rules, not the Docker context.

I built the wheel and confirmed: mcpscanner/docker/Dockerfile and Dockerfile.npm are absent (only entrypoint*.py are included because they're .py). So pip install cisco-ai-mcp-scanner && mcp-scanner pypi-scan flask will fail at docker build with a missing-Dockerfile error i.e. the default, recommended mode is broken for the primary distribution channel. 
Probable Fix: add "mcpscanner.docker" = ["Dockerfile", "Dockerfile.npm"] (and entrypoint*.py, or *) to package-data, and add an install-from-wheel smoke test.

mcpscanner/docker/entrypoint.py
```diff
@@ -0,0 +1,232 @@
...
+
+def download_package(package: str, version: str | None) -> Path:
+    """Download a PyPI package, preferring source dist but falling back to wheel."""
+    os.makedirs(DOWNLOAD_DIR, exist_ok=True)
+
+    spec = f"{package}=={version}" if version else package
+    logger.info("Downloading %s", spec)
+
+    # Try source distribution first
+    cmd_sdist = [
```
sanket-mendapara: PyPI container executes untrusted setup.py as root with the LLM key in its environment.
`pip download` of an sdist runs the package's PEP 517 / setup.py build backend to produce metadata — i.e. arbitrary code execution of the untrusted package. That container is started with the LLM API key in its environment and full outbound network.

So a malicious package can read LLM_API_KEY from its own environment during download and exfiltrate it over the bridge network. Two things stand out:

The npm path already avoids this and it fetches the tarball directly with download_archive() (no npm install, no execution) and runs as a non-root user. The PyPI path could do the same: it already has `download_archive() + safe_extract_archive()` (used by the local mode) and `_resolve_pypi_sdist_url()`. Reusing them in the Docker entrypoint would eliminate code execution entirely and make Docker/local behavior identical.
If execution during download is intentional (to also surface build-time behavior), then at minimum don't pass the LLM key into the download phase and split download and analysis so the secret is only present when parsing, not when executing.

mcpscanner/docker/Dockerfile
```diff
@@ -0,0 +1,9 @@
+FROM python:3.13-alpine
+
+RUN apk add --no-cache gcc musl-dev libffi-dev && \
```
sanket-mendapara: Dockerfile.npm drops to a non-root user (adduser ... USER scanner), but Dockerfile runs as root with no USER.

Neither docker run invocation sets --cap-drop=ALL, --security-opt=no-new-privileges, --pids-limit, --memory, --cpus, or --read-only. This is the container that as mentioned in entrypoint.py file executes untrusted code, so it's the one that most needs hardening. Recommend mirroring the npm image's non-root user and adding cap-drop / no-new-privileges / resource limits to both docker run argvs. (This aligns with the repo's own container-hardening guidance.)

mcpscanner/core/pypi_scanner.py
```diff
@@ -0,0 +1,670 @@
...
+        except json.JSONDecodeError as e:
+            raise PackageDownloadError(
+                f"PyPI returned invalid JSON for {package}: {e}"
+            ) from e
+
+        info = meta.get("info") or {}
+        resolved_version = (info.get("version") or version or "unknown")
+        urls = meta.get("urls") or []
+        sdist = next(
+            (u for u in urls if u.get("packagetype") == "sdist"),
```
sanket-mendapara: Local PyPI mode has no wheel fallback (behavior differs from Docker + docs)

Here, `_resolve_pypi_sdist_url` only accepts packagetype == "sdist" and raises otherwise.
Docker mode falls back to wheels (entrypoint.py:55-64), and both the README and docs/pypi-scanning.md state "falls back to wheels if unavailable." So a wheel-only package succeeds under Docker but fails under --no-docker. Either add a wheel fallback to local mode or document the mode difference.

mcpscanner/docker/entrypoint.py
```diff
@@ -0,0 +1,232 @@
...
+        if not llm_key:
+            # Refuse to declare is_safe=True for an un-analysed package.
+            raise LLMNotConfiguredError(
+                "LLM_API_KEY not provided to the container; refusing to "
+                "report is_safe=True for an un-analysed package"
+            )
+
+        config = Config(
+            llm_provider_api_key=llm_key,
+            llm_model=os.environ.get("LLM_MODEL", "gpt-4o-mini"),
```
sanket-mendapara: Docker entrypoints here has default model gpt-4o-mini, while local mode's _build_config_from_env resolves DEFAULT_LLM_MODEL = gpt-4o (constants.py:119). Is there a reason to use different models at different places for such analysis.

mcpscanner/docker/Dockerfile
```diff
@@ -0,0 +1,9 @@
...
+    pip install --no-cache-dir cisco-ai-mcp-scanner && \
```
sanket-mendapara: The Dockerfile builds the image by installing cisco-ai-mcp-scanner from PyPI while copying in the local entrypoint.

The copied `entrypoint.py` imports modules that are new in this branch (mcpscanner.core.package_sandbox, mcpscanner.core.analyzers.behavioral.js_code_analyzer, and analysis_scan_status from pypi_scanner). Because the image installs the latest published release rather than the source in this branch, I think there are two problems worth resolving before merge:

Release-ordering hazard. Until a PyPI release that contains these modules exists, the built image's entrypoint will fail at runtime with ImportError. That means the Docker path here is the default and recommended mode and can't actually be exercised from this branch yet. Could we confirm how this was tested end-to-end, and note the required release ordering in the PR description?

Version skew between host and container. Even after a release lands, the image will always analyze with whatever version is newest on PyPI, not the version the operator is running on the host. So results can silently differ from the host's code, and the pip install is currently unpinned, so image contents aren't reproducible across builds.

Suggested options (either works):

- Build the image from the local source tree (e.g. copy the project in and pip install .) so the container matches the code under test, or
- Pin the installed version explicitly (cisco-ai-mcp-scanner==<version>) and pin the base image by digest (FROM python:3.13-alpine@sha256:...) for reproducible builds.

sanket-mendapara (changes requested):
**Some other points:**
1) Dockerfile.npm comment is inaccurate: it says it's "Built from a tagged python:3.13-alpine digest ... pins via Docker content-addressed layers," but FROM python:3.13-alpine is a tag, not a digest. Either pin by digest (@sha256:..., which the supply-chain guidance recommends for both images) or fix the comment.

2) Package name not validated before being passed to pip download / argparse. Not a shell-injection risk (argv is a list), but a name beginning with - or containing pip requirement syntax produces confusing failures; a simple PEP 508 name check would be friendlier and tighter.

3) SSRF defense-in-depth: download_archive enforces an HTTPS host allow-list on every redirect hop (good), but doesn't block private/link-local resolution. Given the allow-list + digest verification the residual risk is low; worth a note if you want to be thorough.

4) docs/pypi-scanning.md step 4 implies .whl extraction is part of the standard flow; in local mode only sdists are handled

sanket-mendapara (approved):

PR #156, opened 2026-04-09
Fix/readiness analyzer mcp spec alignment

harishsg999:
removed some rules to match with MCP Spec


n0tjustanychang:
assessed not necessary, closing

PR #158, opened 2026-04-20
fix: surface safe tools in behavioral scan results

harishsg999:
The behavioral scanner previously emitted results only for tools with findings, so benign tools were silently dropped from CLI output and summaries. Track every analyzed function in BehavioralCodeAnalyzer, build a ScanResult per tool (safe or unsafe) in the CLI, and render a dedicated Safe Items section in the summary report. This gives users full visibility into what was scanned, matching the behavior expected for remote/static scans.

### DIFF
--- a/mcpscanner/core/report_generator.py
+++ b/mcpscanner/core/report_generator.py
@@ -578,7 +578,7 @@ class ReportGenerator:
                 "INFO": "🔵",
                 "SAFE": "🟢",
             }
-            severity_icon = severity_emojis.get(highest_severity, "🟢")
+            severity_icon = severity_emojis.get(highest_severity, "🟣")
             output.append(f"{severity_icon} {tool_name} ({highest_severity})")
 
             if total_findings > 0:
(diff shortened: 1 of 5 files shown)


mcpscanner/core/report_generator.py
```diff
@@ -357,44 +357,58 @@ def _format_summary(self, results: List[Dict[str, Any]]) -> str:
...
+            elif item_type == "prompt":
+                item_name = result.get("prompt_name", "Unknown")
+            elif item_type == "resource":
+                item_name = result.get("resource_name", "Unknown")
+            else:
+                item_name = "Unknown"
+
+            findings = result.get("findings", {})
+
+            highest_severity = "SAFE"
```
sanket-mendapara: initially, the highest_severity should be UNKNOWN right!? then after analysis it is either any of UNSAFE or SAFE ones.

sanket-mendapara (changes requested):

sanket-mendapara (approved):

PR #161, opened 2026-05-04
fix: detect critical /etc file paths in YARA rule

alephnull-sh:
## Summary

The `system_manipulation` YARA rule already calls out `/etc/passwd`, `/etc/shadow`, `/etc/hosts`, and `/etc/pwd`, but the regex only matched those filenames when they had a trailing slash.

That meant a normal reference like `cat /etc/passwd` did not hit this rule, while `cat /etc/passwd/` did. This switches that part of the pattern to a word boundary so the file path itself is detected, and adds a regression test for the common `/etc/passwd` / `/etc/shadow` case.

I also added a small guard test for `/etc/passwded` so the rule does not start matching longer words with the same prefix.

## Tests

```bash
.venv/bin/python -m pytest tests/test_yara_rules_coverage.py -q
# 18 passed in 1.23s

.venv/bin/python -m pytest tests/test_yara_analyzer.py -q
# 11 passed in 0.94s

.venv/bin/python -m black --target-version py312 --check tests/test_yara_rules_coverage.py
# 1 file would be left unchanged

git diff --check
```

### DIFF
--- a/mcpscanner/data/yara_rules/system_manipulation.yara
+++ b/mcpscanner/data/yara_rules/system_manipulation.yara
@@ -23,7 +23,7 @@ rule system_manipulation{
         $permission_manipulation = /\b(chmod\s+(777|4755|6755|[ug]?\+s)|(chown|chgrp)\s+(root|0)|setuid|setgid)\b/i
 
         // System directory and file access
-        $critical_system_access = /(\/etc\/(passwd|pwd|shadow|hosts)\/|(\/usr)?\/s?bin\/|\/(tmp|var|root)\/|C:\\\\Windows\\\\System32\b)/i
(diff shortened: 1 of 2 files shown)


sanket-mendapara (approved):
Looks good

alephnull-sh:
@sanket-mendapara thanks for the review.

I think the remaining block is repo-side rather than this patch. All visible CI jobs are green, but the main ruleset still requires a `CodeQL` status check. This PR's head commit does not have a `CodeQL` check run/status, so GitHub is leaving it blocked.

Could you help kick CodeQL for this PR, or route it to whoever can adjust that required check for fork PRs? Happy to change anything in the PR if there is something I can fix from my side.

n0tjustanychang:
@Big-Comfy thanks for flagging the `codeQL` issue

PR #160, opened 2026-04-27
feat(analyzer): add MCPS protocol security analyzer (DAST)

razashariff:
Following up from cisco-ai-defense/defenseclaw#133 per @vineethsai7 request to add this to mcp-scanner instead.

## Summary

Adds `ProtocolAnalyzer` -- a new analyzer that probes live MCP server endpoints for protocol-level security controls via JSON-RPC.

Adds a runtime protocol verification layer alongside existing analyzers (LLM, YARA, API, behavioral) that also connect to MCP server URLs. Where existing analyzers assess server behavior through their respective methods, this analyzer focuses specifically on protocol-level security posture: transport encryption, authentication enforcement, message signing, replay protection, and tool integrity.

## 9 Checks

| ID | Check | Severity | CWE |
|---|---|---|---|
| MCPS-001 | Unencrypted HTTP transport | HIGH | CWE-319 |
| MCPS-002 | Unauthenticated requests accepted | HIGH | CWE-306 |
| MCPS-003 | No message signing detected | HIGH | CWE-345 |
| MCPS-004 | No replay protection | MEDIUM | CWE-294 |
| MCPS-005 | Tool definitions lack integrity hashes | MEDIUM | CWE-494 |
| MCPS-006 | Tool poisoning in descriptions | HIGH | CWE-74 |
| MCPS-007 | Spoofed agent identity accepted | HIGH | CWE-290 |
| MCPS-008 | Fail-open semantics | MEDIUM | CWE-636 |
| MCPS-009 | No rate limiting | MEDIUM | CWE-770 |

## Architecture

- Implements `BaseAnalyzer` interface
- Returns `SecurityFinding` objects with severity, threat_category, CWE, and OWASP references
- Uses `httpx.AsyncClient` for async HTTP (already a project dependency)
- No new external dependencies

## Example Finding Response

\`\`\`json
{
  "severity": "HIGH",
  "summary": "MCP server is accessible over unencrypted HTTP. All JSON-RPC messages, tool arguments, and responses are visible to network observers.",
  "threat_category": "Unencrypted Transport",
  "analyzer": "PROTOCOL",
  "details": {
    "check_id": "MCPS-001",
    "cwe": "CWE-319",
    "target": "http://example.com",
    "owasp_ref": "MCP Security Cheat Sheet Section 6"
  }
}
\`\`\`

## Tests

6 test functions with 3 test servers (vulnerable, secure, poisoned). Covers all 9 checks.

## References

[...]

PR #163, opened 2026-05-05
fix: follow MCP HTTP lifecycle in protocol analyzer

alephnull-sh:
## Summary

Follow-up to #160. The protocol analyzer now does the MCP HTTP handshake before running the probes:

- sends `initialize` first, using the MCP package's latest protocol version
- sends `notifications/initialized` after a successful initialize response
- carries the negotiated `MCP-Protocol-Version` on later requests
- carries `MCP-Session-Id` when a stateful HTTP server returns one

That matters for stateful Streamable HTTP servers. If they return a session ID during initialize, they can reject later `tools/list` requests without that header. In that case the analyzer could miss checks that depend on `tools/list`, replay probes, or rate-limit probes.

I added a small stateful HTTP test server that enforces both headers and checks that the analyzer still reaches the replay, tool integrity, and rate-limit checks.

## Tests

- `.venv/bin/python -m pytest tests/test_protocol_analyzer.py -q`
- `.venv/bin/python -m black --check mcpscanner/core/analyzers/protocol_analyzer.py tests/test_protocol_analyzer.py`
- `.venv/bin/python -m py_compile mcpscanner/core/analyzers/protocol_analyzer.py tests/test_protocol_analyzer.py`


harishsg999 (approved):
lgtm

PR #169, opened 2026-05-08
security: reject symlinks that escape the scan root in directory scans

harishsg999:
Both BehavioralCodeAnalyzer and VirusTotalAnalyzer walked a user-supplied directory with Path.rglob and then opened or hashed each match. rglob matches symlinked files and the subsequent open()/read_bytes() follows them, so a symlink inside the scan root pointing at a sensitive file outside the root caused the external file to be ingested by the analyzer (and, with VT uploads enabled, exfiltrated to a third-party service).

Fix:
- New mcpscanner/utils/path_safety.py with safe_resolve_root / is_within_root / filter_safe_paths helpers. Each candidate is canonicalised and compared against a one-shot resolved root; anything escaping the root is dropped and logged at WARNING (the resolved escape target is only emitted at DEBUG to avoid mirroring sensitive paths into shared logs).
- Behavioral analyzer _find_source_files / _find_python_files route through filter_safe_paths.
- VirusTotal analyzer _discover_files routes through filter_safe_paths before any hashing happens.
- Intra-root symlinks are still allowed (e.g. node_modules/.bin) so scan coverage of legitimate trees is unchanged.

Adds tests/test_path_safety_symlink_traversal.py covering: helper behaviour, both analyzer discovery paths, intra-root symlinks, broken symlinks, escaping directory symlinks, and symlink cycles. All 15 new tests pass; full suite remains green (682 passed, 7 skipped).

### DIFF
--- a/mcpscanner/core/analyzers/behavioral/code_analyzer.py
+++ b/mcpscanner/core/analyzers/behavioral/code_analyzer.py
@@ -33,6 +33,7 @@ from typing import Any, Dict, List, Optional, Union
 from ....config.config import Config
 from ....config.constants import MCPScannerConstants
 from ....threats.threats import ThreatMapping
+from ....utils.path_safety import filter_safe_paths, safe_resolve_root
 from ...static_analysis.context_extractor import ContextExtractor
 from ...static_analysis.native_analyzer import NativeAnalyzer
 from ...static_analysis.interprocedural.call_graph_analyzer import CallGraphAnalyzer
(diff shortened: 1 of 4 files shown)


ablox-chetan (approved):
Security review complete. I focused on the symlink escape path in directory scans, especially whether candidates are filtered before behavioral reads and VirusTotal magic/hash/upload work.

No blocking security issues found. The new helper resolves candidates against a canonical scan root, the analyzer integrations apply it before content reads, and the targeted regression suite passed locally: `uv run pytest tests/test_path_safety_symlink_traversal.py -q` -> 15 passed.

One non-blocking cleanup: in `_discover_files`, the comment says `is_file(follow_symlinks=False)`, but the code calls default-following `is_file()`. The current behavior still rejects escaping symlink targets before content is read and preserves intra-root symlinks, but the comment should be aligned with the actual behavior in this PR or a follow-up.

PR #172, opened 2026-05-18
fix(scanner):  scan_remote_server_resources/prompts fails with "Session terminated" against tools-only MCP servers 

harishsg999:
Servers that don't implement resources/list or prompts/list often reply with a plain HTTP 404 instead of the spec-compliant JSON-RPC -32601 ("Method not found"). The MCP Python SDK
(mcp/client/streamable_http.py) relabels that 404 as a synthetic JSON-RPC error with code 32600 and message "Session terminated", which ClientSession.send_request then raises as McpError.

mcp-scanner's _is_missing_capability_error only recognized -32601 and a handful of message tokens, so all four scan entrypoints (scan_remote_server_resources / _resource / _prompts / _prompt) bubbled this as a 500-style error for very common real-world servers (BigQuery at https://bigquery.googleapis.com/mcp, several Google / Cloudflare / GitHub MCP endpoints). The asymmetric symptom: tools scans succeed, resource/prompt scans fail with the same auth + session.

Two layered fixes, both in mcpscanner/core/scanner.py:

1. _is_missing_capability_error: also treat code == 32600 + "session terminated" (case-insensitive) as a missing-capability signal. The match is intentionally narrow — code AND message — so real mid-session terminations (which can carry 32600 with other messages) still bubble.

2. _server_supports_capability(session, name): read session._init_result.capabilities (already stashed by _get_mcp_session at session.initialize() time) and short-circuit the four scan entrypoints BEFORE the network call when the server didn't advertise the capability. List scans return []; single-item scans raise ValueError so callers can map to 4xx. Robust to any future server quirk that hides the unsupported-method signal in a different way.

Tests (tests/test_scanner.py, +10 cases):
  TestIsMissingCapabilityError
    - spec -32601, synthetic 32600+"Session terminated", case-insensitive
    - critically: 32600 with a different message must NOT match (no silent swallowing of real session-management failures)
    - existing token paths still recognized TestServerSupportsCapability
    - None when session has no _init_result / no .capabilities
    - False when capability was explicitly omitted
    - True for an empty-but-non-None ServerCapabilities sub-model Integration tests for all four scan entrypoints
    - capability gate short-circuits list_resources/list_prompts (asserted via assert_not_called())
    - synthetic Session-terminated at the network layer also handled cleanly (returns [], close session)

### DIFF
--- a/mcpscanner/core/scanner.py
+++ b/mcpscanner/core/scanner.py
@@ -251,7 +251,23 @@ class Scanner:
 
     @staticmethod
     def _is_missing_capability_error(error: Exception) -> bool:
-        """Return True when the server reports a capability is unavailable."""
+        """Return True when the server reports a capability is unavailable.
+
+        Covers three real-world shapes of "this method isn't implemented":
+
+        1. Spec-compliant JSON-RPC ``-32601`` ("Method not found").
+        2. Free-form message tokens (some servers return ``-32603`` with a
+           ``"unsupported"``-style message).
+        3. The MCP Python SDK's synthetic ``32600`` + ``"Session terminated"``
+           that ``mcp/client/streamable_http.py`` emits when the server
+           replies to a JSON-RPC POST with a plain HTTP 404. Many real-world
+           MCP servers (BigQuery, several Google APIs, some Cloudflare/GitHub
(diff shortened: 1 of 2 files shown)


sanket-mendapara (approved):
lgtm. @sisambam can you also review it once?

PR #177, opened 2026-05-22
fix(behavioral): only analyze MCP capabilities, not every helper func…

harishsg999:
The behavioral analyzer was treating every plain helper function in a source file as if it were an MCP capability. Files like ``def _validate(...): ...`` next to ``@mcp.tool() def add(...)`` reported three "tools" instead of one, and the same bug surfaced in JavaScript/TypeScript via ``server.tool()``-style registrations: a single registered handler plus two helpers showed up as three tool rows in CLI and SDK output. This wasted LLM budget on non-capabilities and made the analyzer's results misrepresent the server's actual MCP surface.

The fallbacks in ``BehavioralCodeAnalyzer._analyze_source_code`` previously called ``NativeAnalyzer.extract_all_function_contexts()`` for non-Python files (and as a Python fallback when ``ContextExtractor`` returned nothing), which by design returns *every* function definition in the file. Switching to a capability-aware extractor was the right fix, but the SDKs vary widely in how they register capabilities, so we needed to verify each pattern against the upstream README/quickstart before coding.

Implementation:

* New ``NativeAnalyzer.extract_mcp_capability_contexts()`` that runs two complementary passes per file:

  1. Function-attached annotations / attributes / macros — covers Java (``@Tool`` / ``@McpTool``), C# (``[McpServerTool]``), Rust (``#[tool]``), PHP (``#[Tool]``), Ruby (``# @tool``). Walks each function node's previous siblings (Rust attribute_item, Ruby comment) plus its child ``modifiers`` / ``attribute_list`` wrappers (Java/C#/PHP). Always preserves the wrapper text so the classifier sees the sigil, then drills into individual annotation children for name extraction.

  2. SDK call-site registrations — covers TS SDK v1 ``server.tool('name', schema, handler)``, TS SDK v2 ``server.registerTool('name', config, handler)``, Go SDK ``mcp.AddTool(server, &mcp.Tool{...}, handler)``, Kotlin SDK ``server.addTool(name=...) { request -> ... }``. The handler is the inline function/arrow/lambda or, when none is inline, the LAST identifier (so Go's leading ``server`` receiver isn't misread as the handler). Object-literal arguments are inspected for ``handler``/``execute``/``fn``/``callback`` fields.

  Results are deduped by AST start-byte across both passes, returned
  with a synthetic ``<annotation>.<kind>`` or ``<registration>.<kind>``
  tag in ``decorator_types`` so downstream code can tell how each
  capability was discovered, and class-qualified for languages that
  scope methods inside a class (``CalcService.add`` for Java/Spring AI,
  ``CalcTools.Add`` for C#, ``Calculator.add`` for Rust impl blocks).

* Cross-grammar tree-sitter helpers needed to make the above work: ``_ts_call_method_name`` now falls back to the last ``identifier``/``simple_identifier`` child of a ``navigation_expression`` (Kotlin) or ``selector_expression`` (Go) when no named field exposes the method name; ``_ts_call_arguments_node`` finds the argument list under unnamed ``value_arguments`` / ``argument_list`` children (Kotlin) when no ``arguments`` field is available; ``_ts_is_callee_of_parent`` recognizes the Kotlin trailing-lambda nesting pattern (outer call → inner call → lambda) via structural position rather than node identity, since the tree-sitter binding can hand back fresh wrapper objects.

* ``BehavioralCodeAnalyzer._analyze_source_code`` now calls the new capability extractor in every branch (Python ContextExtractor, Python NativeAnalyzer fallback, JS/TS branch, unknown-language branch). Plain helpers no longer enter the LLM alignment loop and no longer appear as tool rows in the analyzer output.

* Spurious ``<anonymous>`` row from Rust ``impl Type {}`` blocks is filtered out: ``impl_item`` is treated as a class scope (so ``Calculator.add`` is class-qualified) but is excluded from Pass 1's function-annotation walk via a small ``_TS_NON_FUNCTION_NODE_TYPES`` set so the impl wrapper itself isn't misidentified as a tool by the ``#[tool_router]`` macro.

Tests:

* ``tests/static_analysis/test_capability_extraction.py`` — 32 new tests covering all 9 supported languages (Python, JS, TS, Go, Java, Kotlin, C#, Rust, PHP, Ruby) with three buckets:

  - helpers-only files yield ``[]`` (10 tests, parametrized per language)
  - mixed files (1 capability + N helpers) yield exactly the registered capability with the language-appropriate qualified name (10 tests)
  - returned contexts carry the ``<annotation>.tool`` / ``<registration>.tool`` source-kind tag so downstream code can distinguish discovery paths (9 tests)
  - Python-specific decorator metadata preservation (1 test)
  - ``extract_all_function_contexts()`` backward-compat sanity (1 test)
  - unsupported-language graceful fallback (1 test)

* ``uv run pytest --ignore=tests/integration``: 761 passed, 7 skipped, 1 deselected (the deselected test — ``tests/test_protocol_analyzer.py::test_vulnerable_server_findings`` — is a pre-existing failure on ``origin/main`` that's tied to its vulnerable_server fixture, not to anything in this diff; verified by stashing this PR's changes and re-running it on top of clean main, which produces the same MCPS-002 mismatch).


mcpscanner/core/static_analysis/native_analyzer.py
```diff
@@ -263,6 +779,1774 @@ def extract_all_function_contexts(self) -> List[FunctionContext]:
...
+                handler_node,
+                imports,
+                capability=cap_kind,
+                registered_name=reg.get("name"),
+                source_kind=source_kind,
+            )
+
+        return contexts
+
+    def _resolve_cross_file_handler(
```
sanket-mendapara, line 1146: `_resolve_cross_file_handler` picks the first call-graph entry whose key ends with `::<handler_name>`. If two files in the scanned repo define a function with the same name, the wrong one can be analyzed.

Ex - The dict iteration order matches insertion order, which is "whichever file the call-graph builder ingested first." For a TS file like:

```ts
import { addHandler } from "./tools/add";
server.registerTool("add", {...}, addHandler);
```
the importing file already tells us exactly where `addHandler` lives. It is currently ignored, and a sibling `node_modules/foo/add.ts` or `tests/fixtures/add.ts` that defines `addHandler` currently can win the suffix race and have its body analyzed instead.

Concrete failure shape:

- The behavioral analyzer reports the wrong line/file for the handler.
- The LLM analyzes an unrelated implementation, possibly producing false-positive or missed findings.
- Not a security bug, but a reasoning-quality bug that can silently degrade scan results.

## Proposed fix

1. Build an **import map** for the calling file once per `extract_mcp_capability_contexts` call. Each language's import grammar already gives us the source path:
   - TS/JS: `import { x } from "./tools/add"` → `tools/add` (resolve relative to the importing file).
   - Go: `import alias "github.com/.../pkg"`.
   - Python: `from .tools.add import addHandler`.
2. In `_resolve_cross_file_handler`, prefer matches whose file path matches an import target before falling back to the global suffix search.
3. When the import resolver can't disambiguate (no matching import, ambiguous relative path), keep current suffix behavior but log a `DEBUG` warning so the ambiguity is visible.

mcpscanner/core/static_analysis/native_analyzer.py
```diff
@@ -116,6 +116,522 @@ def _get_language_module(lang: str) -> Optional[Any]:
...
+    """Return ``'template'`` for ``*Template`` registrations, else ``None``.
+
+    The capability kind itself is unchanged (a template is still a
+    prompt/resource), but downstream consumers can branch on the
+    ``<registration>.<kind>.template`` tag emitted into ``decorator_types``.
+    """
+    return "template" if "template" in method_name.lower() else None
+
+
+def _python_decorator_capability(name: str) -> Optional[str]:
```
sanket-mendapara, line 381: The TS branch verifies that `X.tool(...)` is being called on a receiver bound to an MCP server (`_collect_mcp_instances` + `_ts_receiver_is_trusted`). The Python branch does not — it classifies any `@<anything>.tool` decorator as an MCP tool because `_python_decorator_capability` drops the receiver entirely.

Fix directions:

1. Add a Python equivalent of `_collect_mcp_instances` that walks the AST and records local names bound to MCP server instances:
   - `mcp = FastMCP("demo")`
   - `server = Server()`
   - `mcp = mcp.server.fastmcp.FastMCP(...)`
   - Function-parameter binding via type annotation: `def f(mcp: FastMCP): ...`
2. Walk imports and accept the SDK module prefixes already defined in `_MCP_SDK_MODULE_PREFIXES["python"]`.
3. Add a `_py_decorator_receiver` helper that returns the receiver name from `name.rsplit(".", 1)[0]` and rejects unless it's in the instance set.
4. Apply the same "loose if empty" tradeoff as TS (`_ts_receiver_is_trusted`) — if no MCP instance is found in the file but the prefilter fired, default to accepting receivers so unusual import patterns still work.

mcpscanner/core/static_analysis/native_analyzer.py
```diff
@@ -116,6 +116,522 @@ def _get_language_module(lang: str) -> Optional[Any]:
...
+# Compiled once: matches an annotation/attribute/macro sigil followed by an
+# identifier we want to inspect. Covers:
+#   @Tool            Java/Spring AI
+#   @McpTool         Spring AI MCP annotations
+#   [McpServerTool]  C# attributes
+#   #[tool]          Rust attribute macros (and `#[mcp::tool]`)
+#   #[Tool]          PHP 8 attributes
+#   # @tool          Ruby comment-style annotation
+import re as _re  # local alias avoids polluting wider module namespace
+_MCP_ANNOTATION_RE = _re.compile(
```
sanket-mendapara, line 236: The compiled annotation-classifier regex `_MCP_ANNOTATION_RE` only treats `::` as a namespace separator before the captured identifier. As a result, fully-qualified annotations like `@org.springframework.ai.Tool` or `[ModelContextProtocol.Server.McpServerTool]` capture the leading segment (`org` / `ModelContextProtocol`) instead of the trailing identifier (`Tool` / `McpServerTool`) and fall out of classification.

Maybe fix direction:

Extend the optional namespace path in `_MCP_ANNOTATION_RE` to walk `.` and `\` separators as well, not just `::`:

```python
_MCP_ANNOTATION_RE = _re.compile(
    r"""
    (?:
        @\#?
      | \#\s*\[
      | \[
      | \#\s*@
    )
    \s*
    (?:[\w]+(?:::|\\|\.))*    # ← walk ::, \\, or . separators
    ([A-Za-z_][\w]*)
    """,
    _re.VERBOSE,
)
```

That single change makes the captured ident always be the *leaf* identifier, which is what `_MCP_ANNOTATION_IDENTIFIERS` keys on. The downstream `_classify_mcp_annotation` namespace reconstruction logic already handles cross-separator namespace recognition correctly.

-> Test before committing though.

mcpscanner/core/static_analysis/native_analyzer.py
```diff
@@ -263,6 +779,1774 @@ def extract_all_function_contexts(self) -> List[FunctionContext]:
...
+        # consulted multiple times for the same file (annotation index,
+        # registration walk, etc.) and one regex pass is enough.
+        cached = getattr(self, "_mcp_prefilter_cache", None)
+        if cached is not None:
+            return cached
+        result = bool(_MCP_PREFILTER_RE.search(self.source_bytes or b""))
+        self._mcp_prefilter_cache = result
+        return result
+
+    def _py_extract_capability_contexts(self) -> List[FunctionContext]:
```
sanket-mendapara: `_py_extract_capability_contexts` documents a graceful fallback that does not exist in the implementation. The docstring promises that if the "lazy path raises", the function will revert to `extract_all_function_contexts`-and-filter. The body has narrow per-block try/excepts but no outer safety net implementing that contract.

The function body uses narrow guards:

- `try: tree = ast.parse(...) except SyntaxError: return []` (line 1173-1176)
- per-function `try: ctx = self._py_extract_function(node, ...) except Exception: continue` (line 1210-1217)
- analogous per-handler try/except for programmatic registrations (line 1249-1256)

There is no outer `try/except` that re-runs the legacy path.

Fix direction:

Replace the trailing paragraph with something accurate:

```
Returns an empty list if the file isn't parseable Python or if no MCP
capabilities are found. Per-function extraction failures are logged at
DEBUG and skipped individually; they do not abort the whole pass.
```

mcpscanner/core/analyzers/behavioral/code_analyzer.py
```diff
@@ -356,6 +378,50 @@ def _find_source_files(self, directory: str) -> List[str]:
 
         return sorted(source_files)
 
+    def _prefilter_capability_files(self, source_files: List[str]) -> List[str]:
```
sanket-mendapara: During a directory scan, every file that passes the prefilter is read from disk three times: once by the prefilter, once by the call-graph builder, and once by `_analyze_file`. On large repos this is wasted I/O that scales linearly with the number of MCP-marker files.

For each accepted file we pay:

1. `open() + f.read()` in `_prefilter_capability_files` (bytes mode).
2. `open() + f.read()` in the call-graph ingest loop (text mode).
3. `open() + f.read()` again in `_analyze_file` (text mode).

Three system calls and three full reads per file. On a repo where the prefilter accepts 50 files of ~50 KB each, that's 100+ reads and ~5+MBs redundantly streamed.

Fix direction:

Cache the bytes through the pipeline. Change `_prefilter_capability_files` to return `List[Tuple[str, bytes]]` (or a small dataclass) instead of `List[str]`:

```python
@dataclass(slots=True)
class _AcceptedFile:
    path: str
    source_bytes: bytes
    source_text: str

def _prefilter_capability_files(self, source_files: List[str]) -> List[_AcceptedFile]:
    accepted = []
    for src_file in source_files:
        with open(src_file, "rb") as f:
            source_bytes = f.read()
        source_text = source_bytes.decode("utf-8", errors="replace")
        analyzer = NativeAnalyzer(source_text, src_file)
        if analyzer._has_mcp_markers():
            accepted.append(_AcceptedFile(src_file, source_bytes, source_text))
    return accepted
```

Then update the call-graph ingest loop and `_analyze_file` to accept a pre-read `source_code` / `source_bytes` argument when one is available, falling back to `open()` for the single-file path (where the prefilter isn't run).

mcpscanner/core/static_analysis/native_analyzer.py
```diff
@@ -263,6 +779,1774 @@ def extract_all_function_contexts(self) -> List[FunctionContext]:
...
+            attr = handler.attr
+            if base in ("self", "cls") and enclosing_cls:
+                node = class_methods.get(enclosing_cls, {}).get(attr)
+                return node, f"{enclosing_cls}.{attr}"
+            return None, f"{base}.{attr}"
+        # Lambda, factory call, subscript, etc. — surface as unresolved
+        # so the LLM still sees that something was registered here.
+        return None, "<unresolved>"
+
+    def _py_iter_programmatic_registrations(
```
sanket-mendapara, line 2019: The TS branch resolves cross-file handler references via the call graph (`_resolve_cross_file_handler`). The Python branch's programmatic-registration path (`_py_iter_programmatic_registrations`) never calls into the cross-file analyzer — it only checks the current file's `functions_by_name` and `class_methods` dicts. Cross-file Python references always emit unresolved stubs even when the call graph could resolve them.

Concrete failure case (the test `test_python_crossfile_module_attr_emits_unresolved_stub` pins the current behavior):

```python
# server.py
from .tools import docs
mcp.tool()(docs.search_agentcore_docs)
```

The TS equivalent (`server.registerTool("add", config, addHandler)` with `addHandler` imported) resolves to the defining file via the call graph. The Python equivalent always emits an unresolved stub because:

- `_py_iter_programmatic_registrations` doesn't accept a `cross_file_analyzer`.
- `_py_resolve_handler_expr` only checks the in-file `functions_by_name` dict.

The unresolved stub still surfaces the capability name, so detection isn't lost — but the body is never analyzed. Cross-file dataflow / docstring alignment for Python programmatic registrations is silently degraded relative to TS.

Fix Direction:

1. Thread `cross_file_analyzer` through `_py_extract_capability_contexts` (already takes it) → `_py_iter_programmatic_registrations` → `_py_resolve_handler_expr`.
2. In `_py_resolve_handler_expr`, fall back to `_resolve_cross_file_handler` when the in-file dicts miss

sanket-mendapara (changes requested):

sanket-mendapara (approved):

PR #191, opened 2026-06-10
Fix/behavioral bedrock prompt shape and validator errors

harishsg999:
## Summary

Bedrock-hosted Anthropic models (Haiku 4.5, Sonnet 4.6) were silently
returning empty `{}` for every alignment check. The orchestrator
conflated that with "verified clean" and `BehavioralCodeAnalyzer`
synthesised SAFE rows for every function — producing false-clean
findings artifacts on the IAM-only-Bedrock Lambda path.

Three root causes, fixed together:

1. **73 KB framework template was concatenated into the user message.**
   Anthropic-on-Bedrock penalises long, instruction-heavy user content
   with a `{}` short-circuit. New `build_prompt_parts` /
   `build_batch_prompt_parts` on `AlignmentPromptBuilder` return
   `(system_template, user_payload)` so the template lives in the
   `system` role and the user message carries only per-function
   evidence (~3 KB instead of 77 KB).

2. **`response_format={"type": "json_object"}` was set for every
   non-Azure provider.** That flag is what triggers the `{}`
   short-circuit on Bedrock Anthropic when paired with a multi-KB
   system prompt. Replace the negative `not startswith("azure/")`
   check with a positive allowlist (`_supports_native_json_mode`)
   that only enables the flag for OpenAI direct (`gpt-*`, `openai/*`,
   `o1*`, `o3*`, `chatgpt-*`). Azure stays excluded (preserves the
   existing contract). JSON output is still enforced via the
   system-prompt instruction.

3. **Empty / `{}` / non-JSON / missing-required-fields responses were
   silently treated as "verified clean".** Added
   `AlignmentVerificationError`; `check_alignment` now raises it on
   validator failures so the existing outer except clause emits the
   same `LLM_UNAVAILABLE_KEY` sentinel that transport errors already
   produced — same `severity="ERROR"` finding pattern from #PRIOR_PR.
   Tightened `validate_batch` to return `None` (force per-function
   fallback) on empty arrays, truncated arrays, items missing
   `mismatch_detected`, or non-dict items, instead of silently padding
   with `mismatch_detected=False`.

Verified live against `bedrock/us.anthropic.claude-haiku-4-5-20251001-v1:0`:
a malicious `fetch_documentation` tool now produces a real
`sev=HIGH DATA EXFILTRATION` THREAT finding (1490 chars of structured
JSON evidence including dataflow trace), instead of the prior
empty-`{}` → synthesised-SAFE result.

## Test plan

- [x] New regression suite `tests/behavioral/test_validator_failure_emits_error.py` (21 cases):
  - Each garbage-response shape (empty, whitespace, `{}`, non-JSON, `[]`, missing `mismatch_detected`, mismatch-without-required-fields) → orchestrator returns sentinel tuple, not `None`
  - Each batch-failure shape (empty array, truncated, items missing required field, non-dict item) → per-function fallback yields one sentinel per function
  - Direct unit tests on `validate_batch` lock the new "no silent padding" policy
  - End-to-end: `BehavioralCodeAnalyzer.analyze` against `'{}'`-returning LLM produces 2 ERROR findings, **0 SAFE rows**
  - Message-shape: `system_prompt` kwarg lands in `messages[0]` (system); `response_format` omitted for `bedrock/...`; `response_format` set for `gpt-4o`
- [x] Existing `tests/behavioral/test_llm_unavailable_does_not_emit_safe.py` updated for the new `build_prompt_parts` API
- [x] Full behavioral suite: `105 passed in 2.22s`
- [x] Full repo suite: `844 passed, 7 skipped` (one live-OpenAI integration test fails on sandbox network, unrelated)
- [x] Live Bedrock Haiku 4.5 probe: real structured JSON response, `sev=HIGH` finding with full evidence

## Notes for reviewer

- This PR includes the prior `LLM_UNAVAILABLE_KEY` sentinel commit
  (`fix/behavioral-no-safe-on-llm-error`). The two changes are
  semantically one fix: transport-failure sentinel + validator-failure
  sentinel use the same `_make_llm_unavailable_result` helper. Easiest
  to review them together.
- No public-API breakage. `build_prompt` and `build_batch_prompt` are
  retained; new `build_prompt_parts` / `build_batch_prompt_parts` are
  additive. `verify_alignment` gains an optional `system_prompt` kwarg.
- Azure response-format behaviour is unchanged (regression test
  `test_request_disables_json_mode_for_azure` still passes).

### DIFF
--- a/mcpscanner/core/analyzers/behavioral/alignment/alignment_llm_client.py
+++ b/mcpscanner/core/analyzers/behavioral/alignment/alignment_llm_client.py
@@ -122,11 +122,36 @@ class AlignmentLLMClient:
                 "AlignmentLLMClient initialized with model: %s", self._model
             )
 
-    async def verify_alignment(self, prompt: str) -> str:
+    # Short security-expert preamble that lives in the system role
+    # alongside the (much larger) framework template. Kept tight on
+    # purpose — Anthropic-on-Bedrock penalises long, instruction-heavy
+    # system messages with empty ``{}`` responses, so the bulk of the
+    # framework guidance is appended via ``system_prompt`` from the
+    # prompt builder rather than being baked in here.
+    _BASE_SYSTEM_PREAMBLE = (
+        "You are a security expert analyzing MCP tools. "
+        "You receive complete dataflow, taint analysis, and code context. "
+        "Analyze if the docstring accurately describes what the code actually does. "
+        "Respond ONLY with valid JSON. Do not include any markdown formatting or code blocks."
+    )
+
+    async def verify_alignment(
+        self, prompt: str, *, system_prompt: Optional[str] = None
+    ) -> str:
         """Send alignment verification prompt to LLM with retry logic.
 
         Args:
-            prompt: Comprehensive prompt with alignment verification evidence
+            prompt: User-role payload with the per-function or per-batch
+                evidence the model should analyse. This should contain
+                ONLY the evidence (delimited untrusted-input block); the
+                framework template belongs in ``system_prompt``.
+            system_prompt: Optional system-role content (typically the
+                73 KB framework template returned by
+                ``AlignmentPromptBuilder.build_prompt_parts``). When
+                provided it is concatenated after the built-in
+                ``_BASE_SYSTEM_PREAMBLE``; when ``None`` the legacy
+                "everything in the user role" shape is preserved for
+                backward compatibility.
 
         Returns:
             LLM response (JSON string)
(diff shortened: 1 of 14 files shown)


mcpscanner/core/analyzers/behavioral/alignment/alignment_llm_client.py
```diff
@@ -151,7 +184,9 @@ async def verify_alignment(self, prompt: str) -> str:
```
ihabler: Before this PR, the prompt carried the whole request and routinely tripped the length threshold. Now the bulk rides in system_prompt, which is computed and logged at DEBUG but never compared, so prompt_length basically never crosses the threshold and the warning has gone silent.

ihabler (changes requested):

PR #206, opened 2026-06-25
fix: preserve all individual findings per analyzer in report output

harishsg999:
Aggregating findings per analyzer in results_to_json kept only the first finding's summary, so analyzers that emit many findings per item (notably the readiness analyzer's HEUR-* rules) lost all but one finding in both --raw and --format detailed output.

- results_to_json: add a detailed_findings list per analyzer that preserves each finding's severity, summary, threat_category/type, full details (rule_id, recommendation, location) and mcp_taxonomy. Existing aggregated fields are unchanged for backward compatibility; --raw now includes every finding automatically.
- _format_detailed: render an "Individual Findings (N)" section listing each finding when an analyzer reports more than one.
- tests: cover JSON preservation and detailed rendering of all findings.

## Summary by CodeRabbit

* **New Features**
  * Detailed reports now include individual findings for each analyzer, making it easier to review every issue separately.
  * JSON output now preserves per-finding details alongside the existing summary data.

* **Bug Fixes**
  * Fixed an issue where multiple findings from the same analyzer could be collapsed into a single view.
  * Improved severity reporting so analyzer-level severity reflects the highest individual finding.

### DIFF
--- a/mcpscanner/core/report_generator.py
+++ b/mcpscanner/core/report_generator.py
@@ -57,6 +57,7 @@ async def results_to_json(scan_results) -> List[Dict[str, Any]]:
                     "threat_names": [],
                     "threat_summary": "No threats detected",
                     "total_findings": 0,
+                    "detailed_findings": [],
                 }
                 summaries_by_analyzer[analyzer_key] = []
 
@@ -69,6 +70,7 @@ async def results_to_json(scan_results) -> List[Dict[str, Any]]:
                     "threat_names": [],
                     "threat_summary": "N/A",
                     "total_findings": 0,
+                    "detailed_findings": [],
                 }
                 summaries_by_analyzer[analyzer] = []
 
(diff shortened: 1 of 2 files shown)


moongci38-oss:
Downstream data point, in case it helps prioritize this.

I maintain a tool that consumes mcp-scanner output and maps findings onto disclosure axes. Five of my axes are currently unreachable, and they all trace back to what this PR fixes.

Verified against main today (`mcpscanner/core/report_generator.py`):

- L82 reads `finding.details.get("threat_type", "unknown")` and appends only that to `threat_names`
- `rule_id` appears 0 times in the entire 914-line file

So readiness rule ids never reach `--raw` consumers. Concretely, these three are produced by the analyzer but invisible downstream:

| rule | what it detects | axis it would fill |
|---|---|---|
| HEUR-015 | missing observability config | logging |
| HEUR-018 | risky keywords in tool name/description | tool_permission |
| HEUR-019 | external indicators with no auth field | auth_oauth |

I confirmed the ids do survive when calling the analyzer modules directly, so the loss is in CLI serialization rather than in the analyzers. Reproduction and the CLI-vs-module comparison table:
https://github.com/moongci38-oss/mcp-disclosure/blob/main/docs/research/2026-08-24-scanner-module-call-poc.md

Without those ids a consumer can't attribute a readiness finding to a rule, so it has to report the axis as "not observable by this scanner" even though the analyzer already did the work.

Happy to test a branch against my fixtures if that's useful — 22 readiness findings (8/7/7 across three tools) from a real scan, plus a live-malicious set.

(Tool is MIT and credits mcp-scanner as the engine: https://github.com/moongci38-oss/mcp-disclosure)

PR #209, opened 2026-07-02
fix: use MCP-registered tool names across languages

harishsg999:
Parse Go mcp.Tool{Name: "..."} composite literals (keyed_element/Name), prefer explicit registration names over handler symbols, and default tools to bare method names instead of class-qualified labels so Python, Go, and Rust reports align (e.g. execute_shell_command).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved capability name resolution across languages and registration styles, including explicit name overrides.
  * Resource, tool, prompt, and function names now display expected values instead of combined or class-qualified variants.
  * Improved handler detection for object and composite literal registrations across supported syntax patterns.
  * Reduced false positives from unrelated configuration, route, and model-catalogue definitions.
  * Improved selection of inline handlers, including direct execution fields and multiple-handler registrations, while ignoring nested decoy handlers and retaining inline handler data.

### DIFF
--- a/mcpscanner/core/static_analysis/native_analyzer.py
+++ b/mcpscanner/core/static_analysis/native_analyzer.py
@@ -268,6 +268,33 @@ def _strip_string_quotes(s: str) -> str:
     return s
 
 
+def _bare_handler_symbol(name: str) -> str:
+    """Return the unqualified handler symbol (``Type.method`` -> ``method``)."""
+    if "." in name:
+        return name.rsplit(".", 1)[-1]
+    return name
+
+
+def _effective_capability_name(
+    registered_name: Optional[str],
+    handler_name: str,
+    capability: str,
+) -> str:
+    """Resolve the MCP-registered capability name for reports and LLM prompts.
+
+    Prefer an explicitly registered name (from ``AddTool``, ``#[tool(name=…)]``,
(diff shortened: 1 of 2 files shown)


sanket-mendapara (changes requested):
Issue: `_ts_extract_handler_from_object()` recursively descends through nested object
containers and accepts the first field named `handler`, `execute`, `fn`, `callback`, or
`run`. It does not require that field to be a direct member of the MCP registration
descriptor, and the selected callback can prevent the real positional handler from winning.

Confirmed probe:

```text
server.tool(
  "real",
  { nested: { handler: () => "safe" } },
  async ({ cmd }) => execSync(cmd),
)
```

The extracted context contained the nested callback's string literal, no function calls, and
`has_subprocess_calls=False`. The runtime handler containing `execSync(cmd)` was not
analyzed.

Impact: a malicious JS/TS MCP package can place a benign decoy callback inside schema or
metadata and make the scanner ignore the executable handler.

Suggested fix: inspect only direct registration-descriptor fields. Traverse parser-specific
wrapper nodes where needed, but do not descend into arbitrary nested application objects.
Explicit positional-handler arguments must take precedence, and ambiguous layouts should
produce unresolved/UNKNOWN coverage.

sanket-mendapara (approved):

PR #228, opened 2026-07-31
fix(scanner): stop YARA from stripping description off the shared tool_data

### LINKED ISSUE #227: Readiness analyzer always sees description-less tools when yara is enabled — HEUR-009 guaranteed false positive, HEUR-012 skewed
## Summary

When a scan includes both the `yara` and `readiness` analyzers (the default combination), the readiness analyzer never sees any tool's description: the yara branch deletes `description` from the shared `tool_data` dict before readiness receives the same dict. As a result, HEUR-009 ("vague description") fires for **every tool of every server scanned with yara enabled**, regardless of actual description quality, and HEUR-012 ("no input validation hints") evaluates against a mutated tool definition.

## Root cause

In `Scanner._analyze_tool` (4.8.2):

- `mcpscanner/core/scanner.py:881` — `tool_data = json.loads(tool_json)` creates a single dict reused by later analyzer branches.
- `mcpscanner/core/scanner.py:917` — the yara parameters branch removes the description before scanning parameters (reasonable for yara: the description was already scanned separately):
  ```python
  if "description" in tool_data:
      del tool_data["description"]
  ```
  This mutates the shared dict rather than a copy.
- `mcpscanner/core/scanner.py:961` — the readiness branch then passes the same, now description-less, dict as the tool definition:
  ```python
  readiness_context = {
      "tool_name": name,
      "content_type": "tool_definition",
      "tool_definition": tool_data,
  }
  ```

`ReadinessAnalyzer._check_vague_description` (HEUR-009) sees `description` missing and reports "Tool has no description" (MEDIUM) — even when the live tool ships a multi-kilobyte description. `_check_no_input_validation_hints` (HEUR-012) and any other check that reads the description also operate on the mutated definition.

## Reproduction

Scan any MCP server whose tools have long, detailed descriptions:

```
mcp-scanner --analyzers yara,readiness --server-url <any server>
```

Every tool is reported with HEUR-009 ("has no description"). Run readiness alone:

```
mcp-scanner --analyzers readiness --server-url <same server>
```

and HEUR-009 no longer fires for tools with adequate descriptions. The finding depends on which *other* analyzer ran, not on the tool.

## Impact

- HEUR-009 is a guaranteed MEDIUM false positive for every tool on every yara+readiness scan — it can never be remediated server-side.
- Readiness scores are systematically deflated for all servers scanned with the default analyzer set.
- Combined with the report format collapsing per-finding detail (#198), the cause is invisible to scanned parties: the report just says the tool "does not specify a timeout" (first finding's summary) with an opaque total.

## Suggested fix

Give the yara parameters branch its own copy instead of mutating the shared dict, e.g. at `scanner.py:915-918`:

```python
yara_tool_data = dict(tool_data)
yara_tool_data.pop("description", None)
tool_json_str = json.dumps(yara_tool_data)
```

(or equivalently, deep-copy `tool_data` before the readiness branch). One-line semantics change; no analyzer logic needs to move.

ppcvote:
Closes #227. Root cause and line numbers are @avoskresensky's diagnosis; I verified all three against current `main` and added the regression test.

## The defect

`_analyze_tool` builds a single `tool_data` dict (`scanner.py:881`) that every analyzer branch shares. The YARA parameters branch deleted `description` from it in place, and the READINESS branch then passed that same dict as `context["tool_definition"]`. Since `ReadinessAnalyzer._parse_tool_definition` prefers the context value over the `content` argument, HEUR-009 saw an empty description and fired for **every tool of every server** scanned with the default yara+readiness set — a finding that depended on which *other* analyzer ran, and that no server operator could remediate.

## The fix

Build YARA's description-free payload as a separate view rather than mutating the shared dict. YARA keeps its original behaviour (parameters scanned without the description, which is already scanned separately just above); every later branch sees the tool as the server declared it.

I kept this to the one-line-scope defect rather than introducing a defensive `deepcopy` at the top of `_analyze_tool`. Happy to go the broader route if you'd prefer an explicit immutability contract for `tool_data` — the shape of this bug (analyzer A silently corrupting analyzer B's input) is the kind that recurs, so a structural guard is defensible. Just didn't want to expand the diff without a maintainer preference.

## Verification

New test `test_yara_does_not_strip_description_from_readiness_input` asserts the isolation directly: it captures what READINESS actually receives and checks the description survives, while confirming YARA still gets its description-free payload with nothing else dropped.

- **Without the fix:** the new test fails (`KeyError` on the description READINESS should have seen).
- **With the fix:** passes.
- **Full file:** `54 passed` (53 existing + this one).

Two pre-existing failures in `tests/test_scanner.py` (`test_scan_resources_returns_empty_on_synthetic_session_terminated`, `test_scan_prompts_returns_empty_on_synthetic_session_terminated`) reproduce identically on a clean `main` checkout in my environment and are unrelated to this change — see the note below, which may be the cause.

## Unrelated observation, flagging rather than fixing

While writing the test I noticed that `pyproject.toml` pins `mcp[cli]>=1.25.0`, so a fresh install currently resolves **mcp 2.0.0**. Under 2.0.0, `MCPTool.model_dump_json()` serialises field names rather than aliases, so `tool_data` carries `input_schema`, not `inputSchema`. That makes the check at `scanner.py:937`

```python
if "inputSchema" in tool_data:
```

always false, so the LLM branch silently omits the parameter schema from its analysis content. I have not confirmed the behaviour under mcp 1.x and it is out of scope here, so I've left it alone — but it looked worth a maintainer's attention, and it may also explain the two pre-existing test failures above. Happy to open a separate issue if useful.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Bug Fixes**
  * Preserved tool descriptions during readiness analysis after YARA analysis runs.
  * Ensured different analysis stages receive the correct, independent tool data.

### DIFF
--- a/mcpscanner/core/scanner.py
+++ b/mcpscanner/core/scanner.py
@@ -912,10 +912,17 @@ class Scanner:
 
             # Run YARA analysis on the tool parameters
             try:
-                # Remove description from the JSON as it is already analyzed
-                if "description" in tool_data:
-                    del tool_data["description"]
-                tool_json_str = json.dumps(tool_data)
+                # Remove description from the JSON as it is already analyzed.
+                # Build a view for YARA instead of mutating tool_data: later
+                # branches (notably READINESS, which reads the tool definition
+                # from context) must see the tool exactly as the server
+                # declared it, not as an earlier analyzer left it.
+                yara_params_data = {
+                    key: value
+                    for key, value in tool_data.items()
+                    if key != "description"
+                }
+                tool_json_str = json.dumps(yara_params_data)
                 yara_params_context = {"tool_name": name, "content_type": "parameters"}
                 yara_params_findings = await self._yara_analyzer.analyze(
                     tool_json_str, yara_params_context
--- a/tests/test_scanner.py
+++ b/tests/test_scanner.py
@@ -16,6 +16,7 @@
 
 # tests/test_scanner.py
 
+import json
 import pytest
 import respx
 from unittest.mock import patch, AsyncMock, MagicMock
@@ -1305,3 +1306,65 @@ async def test_scan_prompts_returns_empty_on_synthetic_session_terminated(config
 
     assert results == []
     session.list_prompts.assert_called_once()
+
+
+# --- Regression: analyzers must not corrupt each other's input (issue #227) ---
+
+
+@pytest.mark.asyncio
+async def test_yara_does_not_strip_description_from_readiness_input():
+    """YARA's parameter scan must not mutate the tool definition READINESS sees.
+
+    The YARA branch drops `description` before scanning parameters. When that
+    deletion happened on the dict shared with later branches, READINESS
+    received a description-less tool definition and HEUR-009 ("no description")
+    fired for every tool of every server scanned with the default analyzer
+    set — a finding that depended on which *other* analyzer ran, and that no
+    server operator could remediate.
+    """
+    from mcp.types import Tool as MCPTool
+
+    tool = MCPTool(
+        name="documented_tool",
+        description="A thoroughly documented tool that lists open pull requests.",
+        inputSchema={
+            "type": "object",
+            "properties": {"repo": {"type": "string"}},
+            "required": ["repo"],
+        },
+    )
+
+    seen = {}
+
+    async def capture_readiness(content, context=None):
+        seen["content"] = content
+        seen["context"] = context
+        return []
+
+    async def noop_yara(content, context=None):
+        seen.setdefault("yara_payloads", []).append(content)
+        return []
+
+    scanner = Scanner(Config(api_key="test_api_key"))
+    scanner._yara_analyzer = MagicMock()
+    scanner._yara_analyzer.analyze = AsyncMock(side_effect=noop_yara)
+    scanner._readiness_analyzer = MagicMock()
+    scanner._readiness_analyzer.analyze = AsyncMock(side_effect=capture_readiness)
+
+    await scanner._analyze_tool(
+        tool, [AnalyzerEnum.YARA, AnalyzerEnum.READINESS]
+    )
+
+    # READINESS resolves the tool definition from context first, so that is
+    # the value that must still carry the description.
+    tool_definition = seen["context"]["tool_definition"]
+    assert tool_definition["description"] == tool.description
+    assert json.loads(seen["content"])["description"] == tool.description
+
+    # YARA still gets its description-free parameter payload, and drops
+    # nothing else. Compared as key sets so this does not depend on whether
+    # the MCP model serialises field names or aliases.
+    params_payload = json.loads(seen["yara_payloads"][-1])
+    full_payload = json.loads(tool.model_dump_json())
+    assert "description" not in params_payload
+    assert set(params_payload) == set(full_payload) - {"description"}


harishsg999 (approved):
lgtm

ppcvote:
Branch updated against `main` — it had fallen two commits behind (#215, #232) and was showing `BEHIND`, which blocks the merge button under this repo's up-to-date requirement. Clean merge, no conflicts, and the diff is unchanged: `mcpscanner/core/scanner.py` plus its test.

Current state after the update, in case it is useful:

```
test (3.11)                   SUCCESS
test (3.12)                   SUCCESS
test (3.13)                   SUCCESS
Documentation Check           SUCCESS
YARA Rules Validation         SUCCESS
Prompt Templates Validation   SUCCESS
MCP Scanner Live Tests        QUEUED   <- never started
```

Everything that can run has run and is green. **MCP Scanner Live Tests has been sitting `QUEUED` rather than failing** — from the outside that looks like the fork-PR approval gate rather than anything wrong with the change, since it never started at all. Flagging it because a queued check is easy to miss: it does not show up as a failure anywhere, it simply never reports, and the PR stays `BLOCKED` with an approving review on it.

Nothing needed from me that I can see, but happy to rebase instead of merge, or split anything out, if either would help.

moongci38-oss:
Independent reproduction on 4.8.3, plus one failure mode I didn't see mentioned in #227.

I hit this from the other direction — I was trying to attribute readiness findings to specific HEUR rules and noticed the rule set changed depending on which *other* analyzers were enabled. Same root cause you both identified (`tool_data` mutated in place at `scanner.py:917`, then handed to readiness as `context["tool_definition"]`).

**Additional observation: description-based rules go silent, not just noisy.**

#227 documents HEUR-009 firing as a guaranteed false positive and HEUR-012 being skewed. On my fixtures the same mutation also produces **false negatives** — `_check_no_authentication_context` (HEUR-019) and `_check_no_idempotency_indication` (HEUR-017) both read `description`, so they stop firing entirely when yara is enabled.

Measured on a 3-tool stdio fixture, `mcpscanner` 4.8.3, varying only the analyzer set:

| analyzers | HEUR-009 | HEUR-017 | HEUR-019 |
|---|---|---|---|
| `readiness` | not fired (correct) | fired | fired |
| `readiness,yara` | fired on 3/3 tools (FP) | gone | gone |

Worth noting for anyone triaging by counts: the **total finding count is identical** in both runs (22). One rule enters as a false positive while two leave as false negatives, so aggregate numbers won't reveal it — which may be part of why this went unnoticed for a while.

**Why the false negative matters more to me than the false positive:** a spurious HEUR-009 is visible noise that an operator can learn to ignore. A missing HEUR-019 is a tool that talks to an external API with no documented auth, silently reported as clean.

After applying this PR's approach locally (building the yara payload as a separate view instead of mutating the shared dict), both analyzer combinations produce identical rule sets:

```
HEUR-001×3, HEUR-003×3, HEUR-006×3, HEUR-013×3,
HEUR-014×3, HEUR-015×3, HEUR-017×2, HEUR-018×1, HEUR-019×1
```

I'm not opening a competing PR — this one already has the fix and a regression test. Just adding the data point in case a second independent reproduction helps move it along. Happy to share the fixture server if it's useful for the test suite.

ppcvote:
Update on the check I flagged in August: `MCP Scanner Live Tests` has run and passed rather than sitting queued, so all nine checks are green now. The branch is level with `main` and the approving review is still on it.

Nothing outstanding from my side. Flagging only because the blocker I raised last time has cleared.

PR #233, opened 2026-08-07
Fix : TS dynamic tool registration extraction for wrappers 

harishsg999:
Resolve safeTool-style wrappers, bound class methods, forEach table expansion, and cross-file endpoint loops so static scans report literal tool names instead of parameter or member-expression stubs.

## Summary by CodeRabbit

* **New Features**
  * Expanded capability detection for wrapper functions, bound methods, class properties, and broader handler patterns.
  * Added support for loop-based and table-driven endpoint registrations.
  * Improved cross-file JavaScript and TypeScript analysis, including imported endpoint aliases.
  * Added safeguards for large directory scans while continuing analysis with available results.
* **Bug Fixes**
  * Reduced unresolved names and description placeholders.
  * Prevented duplicate registrations and excluded internal wrapper details.
* **Tests**
  * Added coverage for wrappers, bound handlers, loops, aliases, and cross-file analysis.

### DIFF
--- a/mcpscanner/core/analyzers/behavioral/js_code_analyzer.py
+++ b/mcpscanner/core/analyzers/behavioral/js_code_analyzer.py
@@ -40,6 +40,9 @@ from typing import Any, Dict, List, Optional
 from ....config.config import Config
 from ....config.constants import MCPScannerConstants
 from ....threats.threats import ThreatMapping
+from ...static_analysis.interprocedural.treesitter_call_graph import (
+    TreeSitterCallGraphAnalyzer,
+)
 from ..base import BaseAnalyzer, SecurityFinding
 from .alignment import AlignmentOrchestrator
 
@@ -55,6 +58,18 @@ _JS_EXTENSIONS = (".js", ".mjs", ".cjs", ".jsx", ".ts", ".mts", ".cts", ".tsx")
 _SKIP_DIRS = frozenset(
     {"node_modules", "dist", "build", "out", ".git", ".next", ".turbo", "coverage"}
 )
+
+# Keep in sync with ``BehavioralCodeAnalyzer._EXT_TO_TS_LANGUAGE`` for JS/TS.
+_EXT_TO_TS_LANGUAGE = {
+    ".js": "javascript",
+    ".jsx": "javascript",
+    ".mjs": "javascript",
+    ".cjs": "javascript",
+    ".ts": "typescript",
+    ".tsx": "typescript",
+    ".mts": "typescript",
+    ".cts": "typescript",
+}
 
 
 class JSBehavioralCodeAnalyzer(BaseAnalyzer):
@@ -127,9 +142,14 @@ class JSBehavioralCodeAnalyzer(BaseAnalyzer):
         self.logger.debug(
             "js behavioural scan root=%s files=%d", directory, len(files)
         )
+        call_graphs = self._build_directory_call_graphs(files)
         findings: List[SecurityFinding] = []
         for path in files:
-            file_findings = await self._analyze_file(path, context)
+            file_context = context.copy()
+            lang = self._language_for_path(path)
+            if lang:
+                file_context["cross_file_analyzer"] = call_graphs.get(lang)
+            file_findings = await self._analyze_file(path, file_context)
             findings.extend(file_findings)
         return findings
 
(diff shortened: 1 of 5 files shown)


mcpscanner/core/static_analysis/native_analyzer.py
```diff
@@ -1147,18 +1182,10 @@ def visit_funcs(node):
...
 
-        # Gap 8 extension: static endpoint tables (literal ``name`` /
-        # ``alias`` fields) including arrays reached via ``for (const tool
-        # of api.endpoints)`` loops (in-file and cross-file).
-        table_names = self._ts_collect_static_endpoint_tool_names(
-            tree.root_node,
-            import_target_map=import_target_map,
-            cross_file_analyzer=cross_file_analyzer,
-        )
         for tool_name in table_names:
```
sanket-mendapara, line 1225: Issue: when an inline registration sits inside a recognized table loop and has no literal
name at the call site, the loop filter drops the registration context. The later table pass
emits one name-only unresolved stub per table entry, with no parameters, calls, source, or
security flags.

Confirmed probe:

```text
tools.forEach(m => {
  server.tool(m.name, m.schema, async ({ command }) => execSync(command));
});
```

The only extracted capability was `danger` tagged `<registration.table>.tool`, with empty
calls and `has_subprocess_calls=False`.

Impact: common table-driven registration removes the actual command-executing handler from
behavioral analysis and creates a direct false negative.

Suggested fix: preserve the loop handler context and associate or clone it for each
statically expanded tool name. Do not replace executable evidence with a name-only stub.

mcpscanner/core/static_analysis/native_analyzer.py
```diff
@@ -3049,18 +3129,244 @@ def _ts_parse_registration_args(
...
 
         return {
             "capability": override_capability or _normalize_capability(capability_method),
             "name": name,
+            "name_ref": name_ref,
             "handler_node": handler_node,
             "handler_name": handler_name,
         }
 
+    def _ts_registration_display_name(self, reg: Dict[str, Any]) -> Optional[str]:
```
sanket-mendapara, line 3197: Issue: `_ts_parse_bind_handler_name()` reduces `this.run.bind(this)` to bare `run`.
`_ts_find_function_def_by_name()` then returns the first same-named method indexed in the
file, without preserving the receiver or enclosing class.

Confirmed probe: an earlier `Safe.run()` and a later `Dangerous.run()` registered through
`this.run.bind(this)` produced `danger (Safe.run)`, no calls, and no eval flag even though
the runtime method called `eval`.

Impact: a same-named benign method in another class can make the scanner analyze the wrong
handler and suppress dangerous behavior.

Suggested fix: preserve the full receiver expression, resolve `this.method` against the
registration call's enclosing class, and reject ambiguous bare matches as unresolved/UNKNOWN.

mcpscanner/core/static_analysis/native_analyzer.py
```diff
@@ -3085,23 +3391,31 @@ def _ts_collect_static_endpoint_tool_names(
...
+    ) -> tuple[List[str], List[tuple[int, int]]]:
         """Collect static MCP tool names from descriptor arrays.
 
         Only arrays reached through ``for (const tool of api.endpoints)``-style
         loops are considered (in-file and cross-file). A blanket walk of every
         array literal would mis-tag HTTP route tables and model catalogues as
         MCP tools.
+
+        Returns ``(tool_names, loop_byte_ranges)`` where each range covers a
+        ``for...of`` / ``forEach`` site that expanded a static table.
```
sanket-mendapara: Issue: any local array passed to `forEach` is expanded when its objects contain a supported
name field plus fields such as `description` or `schema`. The callback is not required to
contain a trusted MCP registration or consume the iteration variable in one.

Confirmed probe: an MCP-containing file with
`models = [{name: "gpt-prod", description: "model config"}]` and
`models.forEach(console.log)` emitted `gpt-prod` as `<registration.table>.tool`.

Impact: ordinary model, UI, job, or configuration tables are reported as MCP tools, inflating
coverage and potentially producing false security findings.

Suggested fix: expand a table only when that specific loop callback/body contains a trusted
MCP registration whose name or handler derives from the iteration binding.

mcpscanner/core/static_analysis/native_analyzer.py
```diff
@@ -2380,6 +2409,12 @@ def visit(node: "Node", parent: Optional["Node"]):
                 )
```
sanket-mendapara, line 2448: Issue: a dotted receiver is trusted when any path component equals a trusted MCP instance
name. Knowing that `server` is an MCP instance therefore makes `attacker.server.tool(...)`
trusted as well.

Confirmed probe: after declaring a real MCP `server`, a separate object call
`attacker.server.tool("ghost", ...)` emitted `ghost` as an MCP tool.

Impact: unrelated DSLs or deliberate decoys can inflate the scanned surface and misattribute
code as MCP-reachable.

Suggested fix: track exact trusted expressions and field assignments such as `this.server`.
Do not use component membership as receiver provenance.

mcpscanner/core/static_analysis/native_analyzer.py
```diff
@@ -3049,18 +3129,244 @@ def _ts_parse_registration_args(
...
+    def _ts_find_wrapper_tool_registrations(
+        self, root: "Node", wrapper_functions: Dict[str, str]
+    ) -> List[Dict[str, Any]]:
+        """Registrations made via custom wrappers like ``safeTool('x', ...)``."""
+        if not wrapper_functions:
+            return []
+
+        func_types = self.FUNCTION_NODE_TYPES.get(self.language, set())
+        registrations: List[Dict[str, Any]] = []
+
```
sanket-mendapara, line 3606: Issue: after a module wrapper such as `safeTool` is identified, every bare call with that
text is treated as a wrapper registration. A parameter or local variable that shadows the
wrapper is not resolved lexically.

Impact: shadowed callbacks can create capabilities that are never registered with the MCP
server.

Suggested fix: resolve the call to its lexical declaration and apply wrapper semantics only
when it binds to the detected wrapper function.

mcpscanner/core/analyzers/behavioral/js_code_analyzer.py
```diff
@@ -190,6 +210,63 @@ def _find_js_files(self, directory: str) -> List[str]:
             out.append(str(path))
         return sorted(out)
 
+    @staticmethod
+    def _language_for_path(file_path: str) -> Optional[str]:
+        """Map a source path to its tree-sitter language id."""
+        return _EXT_TO_TS_LANGUAGE.get(Path(file_path).suffix.lower())
+
```
sanket-mendapara, line 231: Issue: `_build_directory_call_graphs()` eagerly reads, parses, and retains every accepted
JS/TS file in per-language analyzers before normal per-file analysis. The per-file limit is
five times the scanner's normal maximum, but this function has no aggregate file, byte,
AST-node, edge, or time budget. Each file is also read again by `_analyze_file()`.

Impact: an untrusted package or directory containing many individually valid files can drive
high memory and CPU use before any finding is produced, causing timeout or incomplete
coverage.

Suggested fix: enforce aggregate graph budgets, build from MCP-relevant/import-reachable
files, reuse cached source during per-file analysis, and return explicit partial/UNKNOWN
coverage when a limit is reached.

sanket-mendapara (changes requested):

mcpscanner/core/static_analysis/native_analyzer.py
```diff
@@ -3049,18 +3170,528 @@ def _ts_parse_registration_args(
...
+        def visit(node: "Node") -> None:
```
sanket-mendapara, line 3607: Wrapper registration detection skips only parameter shadowing through
`_ts_is_parameter_shadow()`. It does not resolve local variable declarations. A local
`const safeTool = other` can therefore be mistaken for the module-level MCP wrapper.

Latest-head probe:

```typescript
function safeTool(name, schema, handler) {
  return server.tool(name, schema, handler);
}

function caller(other) {
  const safeTool = other;
  safeTool("ghost", {}, () => eval("bad"));
}
```

The analyzer emitted a `ghost` capability with `eval` evidence even though the local function
was not the MCP wrapper.

sanket-mendapara (changes requested):
Call-graph resource controls do not cap AST and edge growth

In files:

- `mcpscanner/core/analyzers/behavioral/js_code_analyzer.py`
- `mcpscanner/core/static_analysis/interprocedural/treesitter_call_graph.py`

The latest commit adds aggregate file and byte limits and releases cached source after each file. That fixes the original unbounded source-cache problem. However, `_find_js_files()` still enumerates an unbounded directory, and the call-graph builder retains parsed trees, functions, and call edges without node, edge, or time budgets. A small number of near-limit files with large or highly connected ASTs can still consume excessive CPU or memory.

PR #234, opened 2026-08-08
feat(cli): add static-source subcommand for no-LLM YARA source scanning

### LINKED ISSUE #170: behavioral: --analyzer-filter yara_analyzer should skip alignment-verification LLM call
## What I observed

`mcp-scanner behavioral <source-path>` requires an LLM provider API key even when `--analyzer-filter yara_analyzer` is set. The error is consistent across attempts:

```bash
mcp-scanner --analyzer-filter yara_analyzer behavioral /path/to/src
# Error during scanning: LLM provider API key is required for alignment verification
```

The README presents the behavioral analyzer as part of a multi-engine offering and lists `yara_analyzer` as one of the engines. In practice `behavioral` blocks on alignment verification regardless of `--analyzer-filter`, so the YARA-only path is not reachable through this subcommand.

## Why this matters

A team that wants pattern-detection in CI without per-PR LLM spend has no way to get it from `behavioral` today. The README's "use all three engines together or independently" framing implies the YARA-only path exists; the CLI behavior says it does not.

## Suggested improvement

When `--analyzer-filter yara_analyzer` is the only filter set (or when an explicit `--no-alignment` flag is added), skip the alignment-verification step and run YARA plus secrets detection only. Emit the YARA findings unmodified.

A documentation-only alternative: add a note to the README that `behavioral` always invokes the LLM regardless of `--analyzer-filter`, and add an explicit `yara-only` subcommand that runs YARA on source code without any LLM dependency.

## Reproduction

```bash
uv tool install --python 3.13 cisco-ai-mcp-scanner
mcp-scanner --analyzer-filter yara_analyzer behavioral /path/to/some/python/src
# Error during scanning: LLM provider API key is required for alignment verification
```

Versions: `cisco-ai-mcp-scanner` (latest on PyPI as of 2026-05-15).

### LINKED ISSUE #171: Add a no-LLM 'static-source' mode that runs YARA on raw source directories
## What I observed

The README markets a feature called "Behavioural Code Scanning: Scan Source code of MCP servers for finding threats." A developer who reads that and wants to scan a feature branch's source code before opening a pull request finds three subcommands that touch source.

`static` reads pre-extracted JSON files (the captured output of a live MCP server's `tools/list`, `prompts/list`, and `resources/list` endpoints):

```
usage: mcp-scanner static [-h] [--tools TOOLS] [--prompts PROMPTS]
                          [--resources RESOURCES] [--mime-types MIME_TYPES]
```

`behavioral` reads source code but blocks on the LLM key (see [companion issue]). `remote` connects over SSE or streamable HTTP to a server that is already up.

None of these match the workflow the README sentence describes: scan source code, no live server, no LLM key.

## Why this matters

The developer is on a feature branch, the server is not yet running, the CI secrets have no LLM provider key, and the company does not want to pay an LLM provider per pull request. The README sentence sets an expectation the CLI does not deliver on.

## Suggested improvement

Two ways to close the gap, equally welcome:

1. Add a `static-source` subcommand (or extend `behavioral` with `--no-alignment`) that runs YARA plus secrets detection over a source directory directly. The detection rules already exist; the missing piece is a CLI path that wires them to a directory input without LLM dependency or a running server.

   ```bash
   mcp-scanner static-source /path/to/src
   # YARA + secrets on .py, .js, .ts files in the directory.
   # No tool listings, no LLM, no running server.
   ```

   This puts `mcp-scanner` on parity with `cisco-ai-defense/skill-scanner`, whose default no-LLM mode works as documented.

2. Update the README's "Behavioural Code Scanning" description to say that source scanning requires either a live server or an LLM provider key, so users do not arrive expecting an offline pattern-only path that the CLI does not offer.

Either resolution removes the surprise. The choice is the maintainer's.

Versions: `cisco-ai-mcp-scanner` (latest on PyPI as of 2026-05-15).

Edneam:
## Context

Closes #171. Closes #170.

Both issues describe the same gap from two angles: the README advertises
"Behavioural Code Scanning: Scan Source code of MCP servers for finding threats"
and lists `yara_analyzer` as an independently usable engine, but there is no way
to run pattern detection over source without an LLM provider key.

Reproduced on `main` (4.8.3):

```
$ mcp-scanner --analyzer-filter yara_analyzer behavioral ./src
Error during scanning: LLM provider API key is required for alignment verification
```

The cause is that `--analyzer-filter` is a **report-time** filter
(`report_generator.py:325`), not a runtime one. `BehavioralCodeAnalyzer`
unconditionally constructs an `AlignmentOrchestrator`, whose `AlignmentLLMClient`
raises when no provider key is configured (`alignment_llm_client.py:120`). So the
filter is applied long after the analyzer that needs the key was already built,
and the YARA-only path advertised in the README is unreachable.

## Description

Adds a `static-source` subcommand — the name suggested in #171 — that runs the
existing YARA rules directly over a source file or directory, with no LLM
provider and no running MCP server.

`static` reads pre-captured `tools/list` JSON; `behavioral` reads source but
requires an LLM key. This fills the remaining cell: reads source, needs neither.

Three things worth flagging for review:

**No LLM SDK on the import path.** Importing `BehavioralCodeAnalyzer` to reuse
its directory walk pulls in `litellm` transitively — I measured ~11s and ~880
modules — which would defeat the point of a no-LLM path. So traversal moved to a
new dependency-light `core/source_discovery.py` that both callers share.
`BehavioralCodeAnalyzer._find_source_files` now delegates to it, and the
extension maps are sourced from it, so the two paths cannot drift on which files
count as source. Behaviour there is unchanged.

(Note: `mcpscanner.cli` still imports `litellm` eagerly via the package
`__init__` chain — ~18s on `main` before this branch. Pre-existing and left
alone, but it does mean `static-source` still pays that startup cost today. Happy
to look at deferring it in a separate PR if useful.)

**Clean files are reported, not omitted.** Every scanned file gets a row,
including matches-free ones (`"is_safe": true`, severity `SAFE`). A report
listing only matches would make "examined and clean" indistinguishable from
"never examined".

**A failed scan is not a pass.** If YARA raises on a file, that row is emitted
with status `error` and severity `ERROR` rather than being dropped or defaulted
to clean, and the rest of the scan continues.

Symlink-escape filtering is preserved — the shared traversal keeps
`filter_safe_paths`, so `static-source` cannot be used to read files outside the
scan root. There is a regression test for this.

## Steps to review

```bash
# Reproduce the gap on main first
mcp-scanner --analyzer-filter yara_analyzer behavioral ./some_src   # errors

# Then on this branch, with no LLM key set at all:
unset MCP_SCANNER_LLM_API_KEY
mcp-scanner static-source ./some_src
mcp-scanner static-source ./some_src/server.py --format raw
mcp-scanner static-source ./some_src --rules-path ./custom_rules/
```

Verified locally: all output formats (`summary`, `raw`, `detailed`, `by_tool`,
`by_analyzer`, `by_severity`, `table`), single-file and directory input, empty
directory, missing path (exits 1 with a clean message), `node_modules` /
`__pycache__` / dot-directory skipping, and symlink-escape rejection.

Tests: 11 new tests in `tests/test_source_scan.py`, all passing. Full suite is
1258 passed / 11 skipped, with 4 failures that also fail on unmodified `main`
(`test_prefilter_zero_marker_file_is_fast`,
`test_scan_instructions_api_endpoint`, and the two
`*_synthetic_session_terminated` scanner tests) plus one collection error in
`test_oauth_example_redirects.py` from a missing `mcp.server.fastmcp` — all
pre-existing and unrelated.

New files are `black`-formatted. I deliberately did not reformat the 48
pre-existing files that `black --check` flags on `main`, to keep this diff
reviewable.

## Checklist

- [x] Tests added for the new behaviour (including the failed-scan and
      symlink-escape cases)
- [x] No change in behaviour for existing scan paths
- [x] README updated: documents `static-source`, and notes that `behavioral`
      always requires an LLM key — the expectation mismatch #171 reported
- [x] No new dependencies

## Summary by CodeRabbit

* **New Features**
  * Added a `static-source` command for YARA-based source scanning without an LLM key or running server.
  * Supports individual files or directories, custom rules, detailed output, verbosity, output files, and CI-friendly raw formats.
  * Clearly reports clean, flagged, missing, oversized, and unreadable files.
  * Excludes dependency, build, hidden, and unsafe symlink paths.

* **Documentation**
  * Added usage examples and clarified the difference between `static-source` scanning and behavioral analysis.

### DIFF
--- a/README.md
+++ b/README.md
@@ -480,7 +480,39 @@ mcp-scanner behavioral /path/to/mcp_server.py --output results.json --format raw
 ```
 
 
+> **Note:** `behavioral` always performs LLM-powered alignment checking, so it
+> requires an LLM provider key. For pattern detection over source code with no
+> LLM provider and no running server, use `static-source` below.
+
 See [Behavioral Scanning Documentation](https://github.com/cisco-ai-defense/mcp-scanner/tree/main/docs/behavioral-scanning.md) for complete technical details.
+
+#### Source Code Scanning Without an LLM (`static-source`)
+
+`static-source` runs the YARA rules directly over a source file or directory.
+It needs no LLM provider key and no running MCP server, which makes it usable
+in CI on a feature branch without per-pull-request LLM spend.
+
+Unlike `static`, which reads pre-captured `tools/list` JSON, this scans raw
+source. Unlike `behavioral`, it performs no alignment checking — it is pattern
+detection only, so it will not catch docstring/behavior mismatches.
+
+```bash
+# Scan a directory of source
+mcp-scanner static-source /path/to/mcp_server_src/
+
+# Scan a single file
+mcp-scanner static-source /path/to/server.py
+
+# Custom YARA rules
+mcp-scanner static-source /path/to/src --rules-path /path/to/rules/
+
+# CI-friendly: machine-readable output
+mcp-scanner static-source /path/to/src --output findings.json --format raw
+```
+
+Every scanned file appears in the output, including clean ones (`"is_safe":
+true`), so a report distinguishes "examined and clean" from "never examined". A
+file whose scan fails is reported with status `error` rather than as clean.
 
 #### PyPI Package Scanning
 
--- a/mcpscanner/cli.py
+++ b/mcpscanner/cli.py
@@ -1255,6 +1255,14 @@ async def main():
         default="summary",
         help="Output format (default: %(default)s)",
     )
+    # Also accepted after the subcommand, since that is the natural order for
+    # `static-source SRC --rules-path RULES`. SUPPRESS keeps the subparser from
+    # overwriting a value already given before the subcommand.
+    p_static_source.add_argument(
+        "--rules-path",
+        default=argparse.SUPPRESS,
+        help="Path to directory containing custom YARA rules",
+    )
 
     # Behavioral subcommand - scan local source code
     p_behavioral = subparsers.add_parser(
@@ -2313,6 +2321,11 @@ async def main():
             # a Config or any LLM-backed analyzer: this path must work with no
             # LLM provider key and no running MCP server.
             from mcpscanner.core.source_scan import scan_source_with_yara
+
+            # YARA is the only analyzer that runs here, so the report must not
+            # credit the default api,yara,llm set — otherwise `--format table`
+            # renders API and LLM as SAFE for analyzers that never ran.
+            selected_analyzers = [AnalyzerEnum.YARA]
 
             try:
                 results = await scan_source_with_yara(
--- a/mcpscanner/cli.py
+++ b/mcpscanner/cli.py
@@ -1064,6 +1064,10 @@ async def main():
   %(prog)s static --prompts prompts.json --analyzers llm                     # Scan prompts file
   %(prog)s static --resources resources.json --analyzers yara                # Scan resources file
   %(prog)s static --tools t.json --prompts p.json --analyzers yara,llm,api   # Scan all three types
+
+  # Source code scanning with no LLM key and no running server:
+  %(prog)s static-source /path/to/src                                        # YARA rules over a source tree
+  %(prog)s static-source /path/to/server.py --format raw -o findings.json    # Machine-readable output for CI
         """,
         formatter_class=argparse.RawDescriptionHelpFormatter,
     )
@@ -1206,6 +1210,47 @@ async def main():
         choices=[
             "raw", "summary", "detailed", "by_tool",
             "by_analyzer", "by_severity", "table",
+        ],
+        default="summary",
+        help="Output format (default: %(default)s)",
+    )
+
+    # static-source subcommand - YARA over raw source, no LLM key required
+    p_static_source = subparsers.add_parser(
+        "static-source",
+        help=(
+            "Scan MCP server source code with YARA rules only "
+            "(no LLM key or running server required)"
+        ),
+    )
+    p_static_source.add_argument(
+        "source_path",
+        help="Path to MCP server source code file or directory",
+    )
+    p_static_source.add_argument(
+        "--output",
+        "-o",
+        help="Save scan results to a file",
+    )
+    p_static_source.add_argument(
+        "--verbose", "-v", action="store_true", help="Print verbose output"
+    )
+    p_static_source.add_argument(
+        "--raw", "-r", action="store_true", help="Print raw JSON output"
+    )
+    p_static_source.add_argument(
+        "--detailed", "-d", action="store_true", help="Show detailed results"
+    )
+    p_static_source.add_argument(
+        "--format",
+        choices=[
+            "raw",
+            "summary",
+            "detailed",
+            "by_tool",
+            "by_analyzer",
+            "by_severity",
+            "table",
         ],
         default="summary",
         help="Output format (default: %(default)s)",
@@ -2263,6 +2308,27 @@ async def main():
                 if args.verbose:
                     print(f"Results saved to {args.output}")
 
+        elif args.cmd == "static-source":
+            # YARA-only scan over raw source. Deliberately does NOT construct
+            # a Config or any LLM-backed analyzer: this path must work with no
+            # LLM provider key and no running MCP server.
+            from mcpscanner.core.source_scan import scan_source_with_yara
+
+            try:
+                results = await scan_source_with_yara(
+                    args.source_path,
+                    rules_dir=args.rules_path,
+                )
+            except FileNotFoundError as e:
+                print(f"Error: {e}", file=sys.stderr)
+                sys.exit(1)
+
+            if args.output:
+                with open(args.output, "w", encoding="utf-8") as f:
+                    json.dump(results, f, indent=2)
+                if args.verbose:
+                    print(f"Results saved to {args.output}")
+
         elif args.cmd == "pypi-scan":
             from mcpscanner.core.pypi_scanner import (
                 DockerNotAvailableError,
@@ -2627,6 +2693,8 @@ async def main():
             server_label = f"vulnerable-package:{args.scan_path}"
         elif hasattr(args, "cmd") and args.cmd == "behavioral":
             server_label = f"behavioral:{args.source_path}"
+        elif hasattr(args, "cmd") and args.cmd == "static-source":
+            server_label = f"static-source:{args.source_path}"
         elif hasattr(args, "cmd") and args.cmd == "pypi-scan":
             pkg_spec = args.package
             if getattr(args, "version", None):
@@ -2758,6 +2826,8 @@ async def main():
             server_label = "well-known-configs"
         elif hasattr(args, "cmd") and args.cmd == "behavioral":
             server_label = f"behavioral:{args.source_path}"
+        elif hasattr(args, "cmd") and args.cmd == "static-source":
+            server_label = f"static-source:{args.source_path}"
         elif hasattr(args, "cmd") and args.cmd == "pypi-scan":
             pkg_spec = args.package
             if getattr(args, "version", None):
--- a/mcpscanner/core/analyzers/behavioral/code_analyzer.py
+++ b/mcpscanner/core/analyzers/behavioral/code_analyzer.py
@@ -39,6 +39,11 @@ from ....config.constants import MCPScannerConstants
 from ....threats.threats import ThreatMapping
 from ....utils.log_format import sanitize_log_value, truncate
 from ....utils.path_safety import filter_safe_paths, safe_resolve_root
+from ...source_discovery import (
+    EXT_TO_TS_LANGUAGE,
+    PYTHON_EXTENSIONS,
+    find_source_files,
+)
 from ...static_analysis.context_extractor import ContextExtractor, FunctionContext
 from ...static_analysis.native_analyzer import NativeAnalyzer
 from ...static_analysis.interprocedural.call_graph_analyzer import CallGraphAnalyzer
@@ -526,25 +531,18 @@ class BehavioralCodeAnalyzer(BaseAnalyzer):
             )
             return []
 
-    _EXT_TO_TS_LANGUAGE = {
-        ".js": "javascript", ".jsx": "javascript", ".mjs": "javascript", ".cjs": "javascript",
-        ".ts": "typescript", ".tsx": "typescript", ".mts": "typescript", ".cts": "typescript",
-        ".go": "go",
-        ".java": "java",
-        ".kt": "kotlin", ".kts": "kotlin",
-        ".cs": "c_sharp",
-        ".rb": "ruby", ".rake": "ruby", ".gemspec": "ruby",
-        ".rs": "rust",
-        ".php": "php", ".phtml": "php",
-    }
+    # Sourced from ``core.source_discovery`` so this analyzer and the no-LLM
+    # YARA source scan cannot drift on which extensions count as source.
+    _EXT_TO_TS_LANGUAGE = EXT_TO_TS_LANGUAGE
 
-    _PYTHON_EXTENSIONS = {".py", ".pyw"}
+    _PYTHON_EXTENSIONS = PYTHON_EXTENSIONS
 
     def _find_source_files(self, directory: str) -> List[str]:
         """Find all supported source files in a directory.
 
-        Supports Python, TypeScript, JavaScript, Go, Java, Kotlin, C#, Ruby,
-        Rust, and PHP files.
+        Delegates to :func:`mcpscanner.core.source_discovery.find_source_files`
+        so this path and the no-LLM YARA source scan agree on which files
+        count as source and on symlink-escape filtering.
 
         Args:
             directory: Directory path to search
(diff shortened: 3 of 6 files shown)


Edneam:
Thanks — all four findings were real and reproduced locally. Fixed in 04d6038.

**1. `--rules-path` after the subcommand** (codex + coderabbit)

Confirmed: `static-source src --rules-path rules` exited 2 with "unrecognized arguments", so the workflow I documented in the README was broken. Registered `--rules-path` on the subparser with `default=argparse.SUPPRESS`.

`SUPPRESS` is load-bearing here. With a plain `default=None` the subparser writes `None` over a value supplied *before* the subcommand, silently discarding it:

```
default=None:   ['--rules-path', '/G', 'static-source', '/src'] -> rules_path=None   # clobbered
SUPPRESS:       ['--rules-path', '/G', 'static-source', '/src'] -> rules_path=/G
```

Covered by a parametrised test asserting both argument orders compile the custom rules (it greps the matched rule name out of `--format raw`, so it fails if the rules are silently ignored rather than merely exiting zero).

**2. `requested_analyzers` credited API and LLM**

Confirmed — `--format table` rendered `API: SAFE` and `LLM: SAFE` for analyzers that never ran. `selected_analyzers` is now reset to `[AnalyzerEnum.YARA]` in the branch, so both render `N/A`:

```
before:  clean.py   SAFE   SAFE   SAFE   SAFE
after:   clean.py   SAFE   N/A    SAFE   N/A
```

**3. Unrenderable `ERROR` severity**

Confirmed: `by_severity` iterates a fixed `severity_order` list, so error rows were dropped from the exact format most likely to be used to review failures.

I went with reusing `UNKNOWN` rather than extending the formatters. `get_highest_severity` already documents `UNKNOWN` as *"not yet analyzed" / "analyzer didn't run"*, which is precisely what a failed scan is, and it is already present in every formatter's ordering, the emoji maps, `SeverityFilter`, and `get_statistics`. Adding `ERROR` would have meant touching five shared rendering paths for a semantic that already exists. Happy to switch to a first-class `ERROR` severity if you'd rather have the distinction — it's a formatter change, not a scan-logic one.

Note `is_safe` is still `False` on these rows, so `UNKNOWN` weakens the display severity but not the pass/fail verdict.

**4. Unreadable files silently omitted** (codex)

This was the worst of the four, and thank you for catching it — it's the same fail-open the PR is meant to argue against. A permission-denied tree reported clean:

```
# before, on a directory where every file is chmod 000
[{"tool_name": "No source files found", "is_safe": true, "findings": {}}]
```

`_read_source_text` now returns `(text, error)` and the caller distinguishes three cases: read succeeded, oversized (deliberate skip), and read failed (error row, `is_safe: False`). Oversized files stay skipped since that's an intentional ceiling, not a failure to gather evidence. Also promoted the read-failure log from `debug` to `warning`.

**Verification:** 1263 passed (16 in `tests/test_source_scan.py`, 5 of them new for these findings). The 4 remaining failures reproduce identically on the base commit `893327c` in a clean worktree — pre-existing and unrelated (one is a timing-sensitive perf test that passes in isolation). `tests/test_oauth_example_redirects.py` fails at collection on base too, missing `mcp.server.fastmcp`.

Edneam:
Both findings from the latest review were real and reproduced locally. Fixed in 4312f15.

**1. Unsanitized exception text in failure logs** (`source_scan.py:76-85`, `210-223`)

Reproduced with a file named `evil file=1 "q"`:

```
raw       -> [Errno 13] Permission denied: '/tmp/.../evil file=1 "q"'
sanitized -> [Errno_13]_Permission_denied:__/tmp/.../evil_file_1__q_
```

The spaces, `=` and quotes survive into a structured `key=value` operator log line, so a crafted filename in a scanned tree can forge fields. Both sites now pass `exc` through `sanitize_log_value`, matching your suggested diff.

One deliberate exception: the `threat_summary` report field keeps the **unsanitized** text. That is machine-readable JSON rather than a `key=value` line, and blanking the real error message there would cost operators the actual reason a file failed. There is a test asserting that distinction, so if you would rather sanitize both, it will fail loudly rather than silently drift.

**2. Table-format test asserted too weakly** (`tests/test_source_scan.py`)

Correct, and worse than described. `assert "SAFE" in row` is substring-true for `UNSAFE`, so it also passed on flagged rows; combined with the unbound `row.count("N/A") == 2`, the test would have passed in the exact inverted case it exists to catch — API reporting `SAFE` while YARA reported `N/A`. Now split per column:

```python
assert (api, yara, llm) == ("N/A", "SAFE", "N/A"), row
```

I used a positional `line.split()` rather than parsing header offsets as suggested. No cell in this table contains a space, so the columns split unambiguously, and it avoids ~35 lines of offset-slicing helper in a test file. If the renderer ever emits a value with a space the split breaks visibly rather than silently, which is the failure mode I want in a test.

**Verification** — each fix was reverted in turn to confirm the new tests actually catch the bug:

- reverting `sanitize_log_value(exc)` → both parametrizations fail showing raw `boom=1 "q"` in the log
- reverting `selected_analyzers = [AnalyzerEnum.YARA]` → table test fails with `API=SAFE` for an analyzer that never ran

18/18 tests in `tests/test_source_scan.py` pass. Also re-ran the six CLI paths end to end (`table`, `by_severity`, unreadable file, missing path, `--rules-path` on both sides of the subcommand). Diff is +87/-10 across two files; `cli.py` is byte-identical to the previous commit.

**Unrelated observation, not touched here:** `tests/static_analysis/test_capability_extraction_perf.py::test_prefilter_zero_marker_file_is_fast` asserts a wall-clock `elapsed < 50` ms. On this machine it measures 50-76 ms both with and without my branch, and cProfile puts all of that time in a single `re.Pattern.search` inside `native_analyzer._has_mcp_markers` — code this PR does not touch. Looks like a machine-speed-dependent threshold rather than anything from these changes. Happy to open a separate issue if useful.

harishsg999:
@Edneam  Thanks for contribution! 
I don't think this PR is required since we are looking for mismatch between implementation and description 
which yara cannot solve 
what's opinion on this?

Edneam:
> @Edneam Thanks for contribution! I don't think this PR is required since we are looking for mismatch between implementation and description which yara cannot solve what's opinion on this?

Fair point, and you're right — YARA can't tell you when an implementation doesn't match its description. I wasn't trying to solve that one. behavioral still needs the LLM key in this branch; I didn't touch the alignment path at all.

What I was after is smaller: there's currently no way to reach the YARA engine on source without a key. --analyzer-filter is applied at report time, so by the time it runs, the analyzer that needs the key has already been built and thrown — that's the error in #170.

The case for still doing something is that someone on a feature branch with no server running and no key in CI has nowhere to go: static only takes pre-generated tool JSON. And to be honest about it, YARA on raw source is noisy — it's a way to narrow down what a human or the LLM path looks at, not an answer on its own.

Happy either way: merge it, or I can redo it as behavioral --no-alignment if one entry point is cleaner, or I'll close it if you'd rather not carry the mode. Just say the word.

harishsg999:
> > @Edneam Thanks for contribution! I don't think this PR is required since we are looking for mismatch between implementation and description which yara cannot solve what's opinion on this?
> 
> Fair point, and you're right — YARA can't tell you when an implementation doesn't match its description. I wasn't trying to solve that one. behavioral still needs the LLM key in this branch; I didn't touch the alignment path at all.
> 
> What I was after is smaller: there's currently no way to reach the YARA engine on source without a key. --analyzer-filter is applied at report time, so by the time it runs, the analyzer that needs the key has already been built and thrown — that's the error in #170.
> 
> The case for still doing something is that someone on a feature branch with no server running and no key in CI has nowhere to go: static only takes pre-generated tool JSON. And to be honest about it, YARA on raw source is noisy — it's a way to narrow down what a human or the LLM path looks at, not an answer on its own.
> 
> Happy either way: merge it, or I can redo it as behavioral --no-alignment if one entry point is cleaner, or I'll close it if you'd rather not carry the mode. Just say the word.


Thanks again! but whole point of having behavioural scanning is to check whether tool description matches with implementation or not but I don't see any point of having pattern detection using yara on code base

Edneam:
> > > @Edneam Thanks for contribution! I don't think this PR is required since we are looking for mismatch between implementation and description which yara cannot solve what's opinion on this?
> > 
> > 
> > Fair point, and you're right — YARA can't tell you when an implementation doesn't match its description. I wasn't trying to solve that one. behavioral still needs the LLM key in this branch; I didn't touch the alignment path at all.
> > What I was after is smaller: there's currently no way to reach the YARA engine on source without a key. --analyzer-filter is applied at report time, so by the time it runs, the analyzer that needs the key has already been built and thrown — that's the error in #170.
> > The case for still doing something is that someone on a feature branch with no server running and no key in CI has nowhere to go: static only takes pre-generated tool JSON. And to be honest about it, YARA on raw source is noisy — it's a way to narrow down what a human or the LLM path looks at, not an answer on its own.
> > Happy either way: merge it, or I can redo it as behavioral --no-alignment if one entry point is cleaner, or I'll close it if you'd rather not carry the mode. Just say the word.
> 
> Thanks again! but whole point of having behavioural scanning is to check whether tool description matches with implementation or not but I don't see any point of having pattern detection using yara on code base

Understood — that's a clear enough answer, and it's your call on scope. Closing this.

Two things I hit while working on it that are unrelated to this feature, if either is useful I'll open them separately:

mcpscanner.cli pulls in litellm eagerly through the package __init__ chain — ~18s and ~880 modules on main, paid by every subcommand including --help.
tests/static_analysis/test_capability_extraction_perf.py::test_prefilter_zero_marker_file_is_fast asserts wall-clock elapsed < 50 ms; measures 50–76 ms here on unmodified main, all of it in one re.Pattern.search in native_analyzer._has_mcp_markers.
Thanks for the time on the review either way.

harishsg999:
> > > > @Edneam Thanks for contribution! I don't think this PR is required since we are looking for mismatch between implementation and description which yara cannot solve what's opinion on this?
> > > 
> > > 
> > > Fair point, and you're right — YARA can't tell you when an implementation doesn't match its description. I wasn't trying to solve that one. behavioral still needs the LLM key in this branch; I didn't touch the alignment path at all.
> > > What I was after is smaller: there's currently no way to reach the YARA engine on source without a key. --analyzer-filter is applied at report time, so by the time it runs, the analyzer that needs the key has already been built and thrown — that's the error in #170.
> > > The case for still doing something is that someone on a feature branch with no server running and no key in CI has nowhere to go: static only takes pre-generated tool JSON. And to be honest about it, YARA on raw source is noisy — it's a way to narrow down what a human or the LLM path looks at, not an answer on its own.
> > > Happy either way: merge it, or I can redo it as behavioral --no-alignment if one entry point is cleaner, or I'll close it if you'd rather not carry the mode. Just say the word.
> > 
> > 
> > Thanks again! but whole point of having behavioural scanning is to check whether tool description matches with implementation or not but I don't see any point of having pattern detection using yara on code base
> 
> Understood — that's a clear enough answer, and it's your call on scope. Closing this.
> 
> Two things I hit while working on it that are unrelated to this feature, if either is useful I'll open them separately:
> 
> mcpscanner.cli pulls in litellm eagerly through the package **init** chain — ~18s and ~880 modules on main, paid by every subcommand including --help. tests/static_analysis/test_capability_extraction_perf.py::test_prefilter_zero_marker_file_is_fast asserts wall-clock elapsed < 50 ms; measures 50–76 ms here on unmodified main, all of it in one re.Pattern.search in native_analyzer._has_mcp_markers. Thanks for the time on the review either way.



can you open them as seperate PR?
thanks

Edneam:
Opened the first one as #236 (lazy `litellm` import).

Before I open the second, a correction on what I told you above — I said the perf test "measures 50–76 ms here on unmodified main." That framing was wrong, and the real behaviour is more specific:

- **Idle machine:** `test_prefilter_zero_marker_file_is_fast` passes. 15 direct measurements of the call under test: min 17.0 ms, median 25.9 ms, max 47.5 ms, 0 over the 50 ms budget. Passes 5/5 under pytest.
- **Loaded machine** (load avg ~5.4 on 12 cores): the same test fails on **unmodified `main`**. A/B, 9 reps per arm, alternating under identical load:

| tree | min | median | max | over 50 ms |
|---|---|---|---|---|
| `main` (unmodified) | 49.8 ms | 69.4–75.7 ms | 89.5 ms | 8–9 / 9 |
| #236 branch | 48.1 ms | 62.6–67.8 ms | 89.0 ms | 8–9 / 9 |

So it isn't a regression from anything — it's a wall-clock threshold that tracks machine load. `cProfile` puts effectively all of it in one `re.Pattern.search` call inside `_has_mcp_markers` (23 ms of the 50 ms budget on an idle box, single regex pass over ~39 kB), so there's no hidden work to remove; the budget is just close enough to the real cost that scheduler noise crosses it.

Whether that's worth changing is your call, and it's why I haven't sent it as a code PR yet. The options I see:

1. **Min-of-N** — time the call a few times, assert on the minimum. Keeps the regression-detection power (a real 10x regression still fails) but is robust to a loaded runner. My preference.
2. Raise the budget — simplest, but any number is arbitrary and the test gets weaker.
3. Mark it opt-in (`@pytest.mark.perf`, deselected by default) — honest about it being a benchmark rather than a correctness test.

Say which you'd prefer and I'll send it. If you'd rather not carry a timing test at all, that's a fine answer too and I'll drop it.

One note relevant to #236's own verification: I ran the full suite excluding this test and got **1262 passed, 11 skipped** (1253 on `main` + the 10 new tests in #236, minus this one deselected). With it included the only failure is this pre-existing timing test, on both `main` and the branch.

PR #236, opened 2026-08-13
perf: import litellm lazily so LLM-free paths don't pay for it

Edneam:
## Summary

`litellm` is imported at module scope in four analyzer modules, all of which are reachable from `mcpscanner/core/analyzers/__init__.py`. Because `mcpscanner/__init__.py` imports that package, **every** importer of `mcpscanner` pays the full `litellm` import — including `mcp-scanner --help` and the subcommands that never call an LLM.

This defers the import to the first actual call, behind a small shared helper (`mcpscanner/utils/lazy_litellm.py`). Call sites keep using `await acompletion(...)` unchanged.

Follow-up to a side observation on #234; opening separately as requested.

## Why four files, not one

Fixing a single module doesn't help — whichever of these loads first pulls `litellm` in for the whole process:

- `core/analyzers/llm_analyzer.py`
- `core/analyzers/meta_analyzer.py`
- `core/analyzers/readiness/llm_judge.py`
- `core/analyzers/behavioral/alignment/alignment_llm_client.py`

`readiness/llm_judge.py` already had the right instinct (`try: from litellm import ... except ImportError`), but a `try/except` still *executes* the import. It now uses `importlib.util.find_spec`, which locates the module without executing it, so `is_available()` / `get_unavailable_reason()` keep their existing semantics at no import cost.

## Measurements

A/B against unmodified `main` (`893327c`) via `git worktree`, same interpreter, alternating runs.

| | `main` | this branch |
|---|---|---|
| modules after `import mcpscanner.cli` | 2552 | **832** |
| peak RSS | 228 MB | **73 MB** |
| `--help` wall clock | 17.7–18.5 s | **4.5–5.2 s** |

`--help` output is byte-identical in both arms (9539 chars, exit 0).

One caveat on the wall-clock column: my machine was under load (~5.0, an unrelated process pinning a core), so the absolute seconds are inflated and shouldn't be read as representative. The ratio held across 5 alternating pairs, and the module-count and RSS figures are load-independent — those are the numbers I'd trust. `litellm`, `openai`, and `tokenizers` are all absent from the import graph after this change.

## Tests

`tests/test_lazy_litellm.py` (10 tests). The import-graph assertions run in a **subprocess** on purpose: once any test in a session touches an LLM analyzer, `litellm` is in the parent's `sys.modules` for good, and an in-process check would pass whether or not the production path is lazy.

Full suite: **1262 passed, 11 skipped** (1253 on `main` + 10 new, and see the note below for the 1 deselected).

One design note worth flagging for review, since it's load-bearing and easy to undo by accident: `acompletion` must stay `async def`. `unittest.mock.patch` picks its replacement class via `iscoroutinefunction` — a coroutine function becomes an `AsyncMock`, anything else a `MagicMock` whose return value can't be awaited. The suite patches these names in bare form (`@patch("...llm_analyzer.acompletion")`) and awaits the result, so a plain `def` wrapper breaks 4 existing tests with `object MagicMock can't be used in 'await' expression`. I hit exactly that on my first attempt; there's a comment in the module and a guard test so the next person doesn't.

## Not touched

`tests/static_analysis/test_capability_extraction_perf.py::test_prefilter_zero_marker_file_is_fast` asserts a wall-clock budget (`elapsed < 50` ms) and fails on my machine under load. I checked whether this branch caused it: both `native_analyzer.py` and that test file are byte-identical to `main`, and A/B under identical load shows `main` failing it too (median 69–76 ms, 8–9 of 9 runs over budget) while this branch measures marginally *faster*. So it's a pre-existing machine-speed-dependent threshold, not a regression here — I've left it alone and deselected it in the run above rather than fix an unrelated test in a perf PR. Happy to open a separate issue if you'd like it made load-independent (e.g. asserting the prefilter short-circuits rather than timing it).

I also deliberately did **not** run `black` across the three analyzer files I edited. They were already not black-clean on `main`, and every hunk the formatter wants is in pre-existing code (Bedrock auth branches, timeout logic) — reformatting would bury a 4-line import change in unrelated churn. My two new files are black-clean.

## Summary by CodeRabbit

- **Improvements**
  - AI-powered scanning components now load the optional AI integration only when needed.
  - The application can start and use non-AI functionality even when the AI integration is unavailable.
  - Improved reporting explains when the AI integration cannot be loaded.
  - AI requests now use more consistent model settings and transient-error handling, improving reliability.
  - Alignment checks support configurable retry limits.

- **Tests**
  - Added coverage confirming deferred loading, availability checks, asynchronous requests, and clear error reporting.

### DIFF
--- a/mcpscanner/core/analyzers/behavioral/alignment/alignment_llm_client.py
+++ b/mcpscanner/core/analyzers/behavioral/alignment/alignment_llm_client.py
@@ -31,7 +31,7 @@ import itertools
 import logging
 import time
 
-from litellm import acompletion
+from .....utils.lazy_litellm import acompletion
 
 from .....config.config import Config
 from .....config.constants import MCPScannerConstants
--- a/mcpscanner/core/analyzers/llm_analyzer.py
+++ b/mcpscanner/core/analyzers/llm_analyzer.py
@@ -24,7 +24,8 @@ import asyncio
 import json
 import secrets
 from typing import Any, Dict, List, Optional
-from litellm import acompletion
+
+from ...utils.lazy_litellm import acompletion
 
 from ...config.config import Config
 from ...config.constants import MCPScannerConstants
--- a/mcpscanner/core/analyzers/meta_analyzer.py
+++ b/mcpscanner/core/analyzers/meta_analyzer.py
@@ -45,7 +45,7 @@ import secrets
 from dataclasses import dataclass, field
 from typing import Any, ClassVar, Dict, List, Optional, Tuple
 
-from litellm import acompletion
+from ...utils.lazy_litellm import acompletion
 
 from ...config.config import Config
 from ...config.constants import MCPScannerConstants
--- a/mcpscanner/core/analyzers/readiness/llm_judge.py
+++ b/mcpscanner/core/analyzers/readiness/llm_judge.py
@@ -36,20 +36,13 @@ import os
 from typing import Any, Dict, List, Optional
 
 from ....config.constants import MCPScannerConstants
+from ....utils.lazy_litellm import (
+    acompletion,
+    is_litellm_available as _litellm_is_available,
+    litellm_import_error as _litellm_error_reason,
+)
 from ....utils.logging_config import get_logger
 from ..base import SecurityFinding
-
-# Try to import litellm
-_litellm_available = False
-_litellm_import_error: Optional[str] = None
-
-try:
-    from litellm import acompletion
-
-    _litellm_available = True
-except ImportError as e:
-    _litellm_import_error = str(e)
-    acompletion = None  # type: ignore
 
 
 # Readiness-specific threat categories
@@ -140,7 +133,7 @@ class ReadinessLLMJudge:
         Returns:
             True if the LLM judge can be used, False otherwise.
         """
-        if not _litellm_available:
+        if not _litellm_is_available():
             return False
 
         # Require API key to be set
@@ -153,8 +146,8 @@ class ReadinessLLMJudge:
         Returns:
             Human-readable explanation or None if available.
         """
-        if not _litellm_available:
-            return f"litellm not installed: {_litellm_import_error}"
+        if not _litellm_is_available():
+            return f"litellm not installed: {_litellm_error_reason()}"
 
         if not self.api_key:
             return (
--- a/mcpscanner/utils/lazy_litellm.py
+++ b/mcpscanner/utils/lazy_litellm.py
@@ -1,4 +1,4 @@
-# Copyright 2025 Cisco Systems, Inc. and its affiliates
+# Copyright 2026 Cisco Systems, Inc. and its affiliates
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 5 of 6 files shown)


Edneam:
@coderabbitai Checked this one against the repo's actual tooling before changing anything, and I don't think the suppression is warranted here. Happy to add it if you disagree on any of the below.

**1. There is no ruff gate in this repo, so the finding's stated consequence doesn't apply.**

The comment's rationale is "this prevents the new test file from failing lint." Nothing in the repo runs ruff:

- `pyproject.toml` — no `[tool.ruff]`, and `ruff` is not in `[dependency-groups] dev` (that's `pytest`, `pytest-asyncio`, `pytest-cov`, `respx`, `black`, `flake8`, `isort`, `mypy`, `pre-commit`).
- No `ruff.toml` / `.ruff.toml` anywhere.
- `.pre-commit-config.yaml` — hooks are `check-yaml`, `end-of-file-fixer`, `trailing-whitespace`, `addlicense`, `black`. No ruff.
- CI (`.github/workflows/python-tests.yml`) runs `uv sync --group dev` then `uv run pytest`. No lint step.
- `ruff` isn't even installed in the dev env here (`uv run ruff --version` → not found).

I had to fetch ruff ad-hoc (`uvx ruff@0.16.1`) just to reproduce your finding.

**2. It's one call site, not two.**

`--select S603` reports exactly one:

```
S603 `subprocess` call: check for execution of untrusted input
  --> tests/test_lazy_litellm.py:45:12
```

The second call site you flagged (the availability probe, now line ~117) is **not** flagged by S603. So the requested change was partly based on a finding ruff doesn't actually raise.

**3. Suppressing here would be inconsistent with the rest of the repo.**

Ruff currently reports 7 unsuppressed S603 in `mcpscanner/` production code, and `tests/` has ~2,599 S101 (`assert` in tests) with no suppressions. Adding `# noqa: S603` to one new test file — for a linter the project doesn't run — makes my file the odd one out rather than bringing it in line with a convention.

On the substance of the rule itself: both calls are `sys.executable` with an argument vector (no `shell=True`) and a string literal built in-test, so there's no untrusted input. That's why I'd rather leave the code as-is than annotate around a check that isn't enforced.

**If the plan is to adopt ruff**, the right move is a repo-wide `[tool.ruff.lint.per-file-ignores]` entry (`"tests/*" = ["S101"]`, plus a decision on S603) in a dedicated PR, rather than one-off `noqa`s landing in whichever file a bot reviews first. Glad to open that separately if a maintainer wants it.

One thing your review did prompt me to fix: `mcpscanner/utils/lazy_litellm.py` had a `Copyright 2025` header copied from an older file, while other files added this year (`mcpscanner/utils/path_safety.py`, `tests/static_analysis/test_capability_extraction_perf.py`) use 2026. Corrected to 2026 — pushed.

harishsg999:
@Edneam  can you please resolve the merge conflicts

Edneam:
> @Edneam can you please resolve the merge conflicts

i have resolved the merge conflicts .. please check now

harishsg999:
Thanks 
Can you make this does not break anything and confirm by testing against servers?

Edneam:
> Thanks 
> 
> Can you make this does not break anything and confirm by testing against servers?

I initially tested it , i will test again tomorrow and let you know

PR #238, opened 2026-08-14
Feature/analyzer transient error retry

harishsg999:
Classify and handle transient error with retry

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved analyzer reliability by distinguishing temporary failures from non-retryable errors.
  * Added capped exponential backoff for temporary failures.
  * Authentication and configuration errors now fail immediately without unnecessary retries.
  * Batch analysis retries invalid responses before falling back to individual analysis.
  * Improved handling of oversized prompts and clearer failure reporting, including temporary and final error counts.

* **Tests**
  * Added coverage for error classification, retry behavior, backoff limits, prompt handling, and batch fallback scenarios.

### DIFF
--- a/mcpscanner/utils/analyzer_errors.py
+++ b/mcpscanner/utils/analyzer_errors.py
@@ -25,6 +25,7 @@ without amplifying load against an endpoint that cannot succeed.
 from __future__ import annotations
 
 import asyncio
+import re
 from enum import Enum
 from typing import Awaitable, Callable, Optional, TypeVar
 
@@ -43,9 +44,10 @@ ERROR_KIND_TRANSIENT = ErrorKind.TRANSIENT.value
 ERROR_KIND_FINAL = ErrorKind.FINAL.value
 
 
+_FINAL_STATUS_CODES = frozenset({401, 403})
+_TRANSIENT_STATUS_CODES = frozenset({408, 429, 500, 502, 503, 504})
+
 _FINAL_KEYWORDS = (
-    "401",
-    "403",
     "unauthorized",
     "forbidden",
     "access denied",
@@ -90,11 +92,6 @@ _TRANSIENT_KEYWORDS = (
     "throttl",
     "quota",
     "too many requests",
-    "429",
-    "502",
-    "503",
-    "504",
-    "408",
     "service unavailable",
     "bad gateway",
     "gateway timeout",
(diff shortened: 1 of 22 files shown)


sanket-mendapara:
@harishsg999 -> **GPT-5 request-shape fix only reaches 2 of 4 LLM call sites**
Both meta_analyzer.py (self._model = config.llm_model) and readiness/llm_judge.py (self.model = ...) read from the same configurable model setting as the two fixed call sites — there's no structural reason a user configuring a GPT-5 model would only hit llm_analyzer/alignment_llm_client and not meta-analysis or readiness evaluation.

drop_params=True (present on all four call sites) does not substitute for apply_model_constraints: litellm's drop_params silently drops a parameter the provider doesn't accept — it does not rename max_tokens to max_completion_tokens. So for a GPT-5 model, meta_analyzer.py and readiness/llm_judge.py would have max_tokens silently dropped with no substitute, losing output-length control entirely on every request.

Failure scenario: An operator sets MCP_SCANNER_LLM_MODEL=openai/gpt-5.x globally. LLM analysis and alignment checks work correctly. Meta-analysis (false-positive filtering) and readiness evaluation silently lose their token-length cap on every request — nothing raises or logs a warning about it.

Suggested fix: call apply_model_constraints(self._model, request_params) / apply_model_constraints(self.model, request_params) in meta_analyzer.py:497 and readiness/llm_judge.py:236 before their respective acompletion calls.

mcpscanner/utils/analyzer_errors.py
```diff
@@ -0,0 +1,331 @@
...
+    return ErrorKind.FINAL
+
+
+def compute_backoff_delay(attempt: int, base_delay: float, *, cap: float = 60.0) -> float:
+    """Exponential backoff delay for attempt index ``attempt`` (0-based)."""
+    delay = base_delay * (2**attempt)
+    return min(delay, cap)
+
+
+async def retry_transient_async(
```
sanket-mendapara, line 270: **This retry_transient_async() helper remains unused (dead code)**

Carried over from the previous review round, still unresolved. Grep confirms it's referenced only from tests/test_analyzer_errors.py; all four retry loops (alignment_llm_client.verify_alignment, alignment_orchestrator.check_alignment_batch's parse loop, llm_analyzer._make_llm_request, meta_analyzer._make_llm_request) still hand-roll their own classify → backoff → sleep loop instead of calling it.

Reason: a future fix to retry semantics (jitter, an off-by-one, a new terminal-error condition) must be applied by hand in 4 separate places instead of one.

Suggested fix: either adopt retry_transient_async() at the four call sites, or delete/remove it if it's not meant to be adopted.

sanket-mendapara (changes requested):

sanket-mendapara (approved):

PR #239, opened 2026-08-17
docs: rename "LLM-as-a-judge" to "LLM-based analysis" in README

sisambam:
The README described one of the scanning engines as "LLM-as-a-judge." That term has a specific, established meaning: using an LLM to evaluate or score the output of another system against some criteria. The engine in question does not do that — it reads raw MCP tool metadata and classifies it as malicious or safe, i.e. LLM-based detection/analysis.

By the standard definition, the component that actually fits "LLM-as-a-judge" is the meta-analyzer, which reviews the other analyzers' findings. To avoid confusing readers now that the project is open source, this renames the engine to "LLM-based analysis" in the two README references. Docs-only change; no code, class, enum, or CLI-flag names are affected (the code already calls it LLMAnalyzer).

## Summary by CodeRabbit

* **Documentation**
  * Updated the README overview to describe LLM-based analysis using clearer terminology.


harishsg999 (approved):
lgtm

